PR #2, opened 2016-06-11
Try to use `clang++`

jakirkham:
This is an attempt to make sure `clang++` and `libc++` are used to build SciPy. As there are some C++ STL types being exposed and the rest of our stack using `clang++` and `libc++`, it is very important for SciPy to be compliant. However, this may be tricky given the requirement to have `gfortran`. Let's see what we can do.

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -6,11 +6,31 @@
 # but it works fine to get our first build.
 unset LDFLAGS
 
+# FIXME: This is a hack to make sure the environment is activated.
+# The reason this is required is due to the conda-build issue
+# mentioned below.
+#
+# https://github.com/conda/conda-build/issues/910
+#
+source activate "${CONDA_DEFAULT_ENV}"
+
+
+export LIBRARY_PATH="${PREFIX}/lib"
+export C_INCLUDE_PATH="${PREFIX}/include"
+export CPLUS_INCLUDE_PATH="${PREFIX}/include"
+
 # Depending on our platform, shared libraries end with either .so or .dylib
 if [[ `uname` == 'Darwin' ]]; then
+    # Also, included a workaround so that `-stdlib=c++` doesn't go to
+    # `gfortran` and cause problems.
+    #
+    # https://github.com/conda-forge/toolchain-feedstock/pull/8
     DYLIB_EXT=dylib
+    export CFLAGS="${CFLAGS} -stdlib=libc++ -lc++"
+    export LDFLAGS="-headerpad_max_install_names -undefined dynamic_lookup -bundle -Wl,-search_paths_first -lc++"
 else
     DYLIB_EXT=so
+    unset LDFLAGS
 fi
 
 # If OpenBLAS is being used, we should be able to find the libraries.
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -13,7 +13,7 @@ source:
   md5: 8987b9a3e3cd79218a0a423b21c8e4de
 
 build:
-  number: 200
+  number: 201
   # We lack openblas on Windows, and therefore can't build scipy there either currently.
   skip: true  # [win]
   features:
@@ -21,6 +21,7 @@ build:
 
 requirements:
   build:
+    - toolchain
     - gcc
     - python
     - cython


rgommers:
This failed on the import tests.

jakirkham:
> This failed on the import tests.

Yeah, I saw that last night and decided the best way to sort it out was run it locally and test it out.

> Just checking: we don't care here about OS X versions older than 10.9 right? IIRC those have ``libstdc++`` as default.

10.6 and up have `libc++`. That being said, Travis' copy of XCode appears to only support 10.9. There are a number of things we could do to get 10.7 support (which still meets are requirements of C++11 support), but we are not quite sure how well any of them will work. Given nobody has complained about 10.9 support and everyone thinks we should keep 10.9, we have stuck with it. ( https://github.com/conda-forge/toolchain-feedstock/issues/7 ) Though it sounds like we are going to switch to a Continuum built copy of `gcc` and start shipping `libgcc` as a dependency.

jakirkham:
It looks like some things are getting linked against the Python Framework on the system, which is a little disconcerting.

jakirkham:
Alright, I think I may have gotten this fixed. It at least imports everything ok on my machine. Pushed the latest changes to GitHub.

jakirkham:
Just as FYI, the crashes during import appear to have been caused by `scipy/special/_ufuncs.so` being linked to the system framework Python. It now appears not to linked to any Python and the crashed don't occur. It was determined this library was the problem by noting the crash following the import of `scipy.integrate.vode`. Subsequent running of `python -v -c "import scipy.integrate.vode"` showed the last library passed to `dlopen` before the crash was `scipy/special/_ufuncs.so`. Running `conda inspect linkages -n _test scipy` locally post-build now shows we are now clean of all linkages to the system framework Python.

jakirkham:
Looks like we are running into issues due to the rpath header limit. Will correct momentarily.

jakirkham:
Please feel free to double check what I'm doing here, @rgommers. While I fully expect this will pass on all platforms, there are probably things that still need tweaking.

recipe/build.sh
```diff
@@ -6,11 +6,32 @@
...
+# The reason this is required is due to the conda-build issue
+# mentioned below.
+#
+# https://github.com/conda/conda-build/issues/910
+#
+# Also, included a workaround so that `-stdlib=c++` doesn't go to
+# `gfortran` and cause problems.
+#
+# https://github.com/conda-forge/toolchain-feedstock/pull/8
+OLD_LDFLAGS="${LDFLAGS}"
```
rgommers: Is this some kind of future-proofing? ``LDFLAGS`` is unset above, so this isn't doing anything now

rgommers:
The set of LDFLAGS looks fine to me - except for ``headerpad_max_install_names`` and ``-lc++`` they're the same as we used to build Scipy dmg installers with

recipe/build.sh
jakirkham: You're right. I'll remove this.

jakirkham:
We had to reset the `LDFLAGS` here because the `-stdlib` argument was going to `gfortran`, which errored out when receiving it. I have PR ( https://github.com/conda-forge/toolchain-feedstock/pull/8 ) to change that in the `toolchain`.

Adding `-headerpad_max_install_names` is something the `toolchain` package would normally do. This is of particular value with `conda-build` as it needs a way to change the rpath after a build so that binaries can be relocated easily. However, it can sometimes run out of space in the header thus failing to change the rpath. Setting the flag `-headerpad_max_install_names` makes sure that each path is given enough space (probably through null characters) to fill up to `MAXPATHLEN`, which is 1024 on Mac. ( [ref]( https://developer.apple.com/library/ios/documentation/System/Conceptual/ManPages_iPhoneOS/man2/statfs64.2.html ) )

The reason we are setting `-lc++` is we are trying to ensure that we are definitely linked to `libc++` for C++ support instead of `stdlibc++`. As we use `libc++` throughout the rest of our stack (as we need C++11 support for things), this ensures that linking against SciPy's C++ interface should be ok. Though I'm unaware to what extent that is done.

Both of these added flags are already in the `toolchain` package. Once the aforementioned PR ( https://github.com/conda-forge/toolchain-feedstock/pull/8 ) is accepted, we can clean this up a bit more.

Does this seem reasonable? More thoughts/questions?

rgommers:
> As we use libc++ throughout the rest of our stack (as we need C++11 support for things), this ensures that linking against SciPy's C++ interface should be ok. Though I'm unaware to what extent that is done.

It's not. Scipy exposes Cython interfaces to BLAS and LAPACK, and soon also to many functions in ``scipy.special``, but no C++ ones.

recipe/meta.yaml
```diff
@@ -13,15 +13,17 @@ source:
...
   # We lack openblas on Windows, and therefore can't build scipy there either currently.
   skip: true  # [win]
   features:
     - blas_{{ variant }}
 
 requirements:
   build:
+    - toolchain
     - gcc
+    - cloog 0.18.0 10
```
rgommers: I looked at cloog.org, but I don't see how this is related to Scipy?

jakirkham: Yeah, it's not. It's related to the fact that `gcc` package does not pin `isl` correctly. As we have a newer version of `isl` than `defaults`, this causes problems with the compiler. We have tried to rebuild the compiler, but do not have sufficient build time ATM.

We came up with a fragile workaround that used to work by default for awhile using this `cloog` package. However, it seems changes in the solver have caused that to fail. Basically this `cloog` package (which we build [here]( https://github.com/conda-forge/cloog-feedstock )) takes care of pinning `isl` correctly as `gcc` does not. To get a better idea of what is going on please take a look at this issue ( https://github.com/conda-forge/staged-recipes/issues/436 ) and this issue ( https://github.com/conda-forge/isl-feedstock/issues/3 ). It looks like Continuum will rebuild the `gcc` package, but we don't know the time frame of that change yet. So, this is simply the explicit form the previously implicit workaround. As it is explicit, it is not fragile in the same way.

rgommers: Thanks @jakirkham. Maybe add a comment in the recipe explaining the purpose of ``cloog``?

jakirkham: Sure, sounds reasonable.

jakirkham: `gcc` got hot-fixed. So, I think we can drop this.

jakirkham:
Alright, I think I have addressed everything. Please let me know if there are more comments.

rgommers:
Allright, all looks good so let's give it a go I'd say.

rgommers:
Merged, thanks @jakirkham!

jakirkham:
Thanks @rgommers.

jakirkham:
Despite all builds passing, it seems some binaries from Travis CI were not uploaded. Will restart them selectively. This is the error one such build shows.

```bash
[32;1mThe command "conda build ./recipe" exited with 0.[0m

travis_fold:start:after_success
[0Ktravis_time:start:0e2e2414
[0K$ ./ci_support/upload_or_check_non_existence.py ./recipe conda-forge --channel=main

Using Anaconda Cloud api site https://api.anaconda.org

Using Anaconda Cloud api site https://api.anaconda.org

[91m[ERROR] [0m

  File "/Users/travis/miniconda3/lib/python3.5/site-packages/requests/packages/urllib3/connectionpool.py", line 376, in _make_request

    httplib_response = conn.getresponse(buffering=True)

TypeError: getresponse() got an unexpected keyword argument 'buffering'



During handling of the above exception, another exception occurred:



  File "/Users/travis/miniconda3/lib/python3.5/site-packages/requests/packages/urllib3/connectionpool.py", line 559, in urlopen

    body=body, headers=headers)

  File "/Users/travis/miniconda3/lib/python3.5/site-packages/requests/packages/urllib3/connectionpool.py", line 378, in _make_request

    httplib_response = conn.getresponse()

  File "/Users/travis/miniconda3/lib/python3.5/http/client.py", line 1197, in getresponse

    response.begin()

  File "/Users/travis/miniconda3/lib/python3.5/http/client.py", line 297, in begin

    version, status, reason = self._read_status()

  File "/Users/travis/miniconda3/lib/python3.5/http/client.py", line 266, in _read_status

    raise RemoteDisconnected("Remote end closed connection without"

http.client.RemoteDisconnected: Remote end closed connection without response



During handling of the above exception, another exception occurred:



  File "/Users/travis/miniconda3/lib/python3.5/site-packages/requests/adapters.py", line 376, in send

    timeout=timeout

  File "/Users/travis/miniconda3/lib/python3.5/site-packages/requests/packages/urllib3/connectionpool.py", line 609, in urlopen

    _stacktrace=sys.exc_info()[2])

  File "/Users/travis/miniconda3/lib/python3.5/site-packages/requests/packages/urllib3/util/retry.py", line 247, in increment

    raise six.reraise(type(error), error, _stacktrace)

  File "/Users/travis/miniconda3/lib/python3.5/site-packages/requests/packages/urllib3/packages/six.py", line 309, in reraise

    raise value.with_traceback(tb)

  File "/Users/travis/miniconda3/lib/python3.5/site-packages/requests/packages/urllib3/connectionpool.py", line 559, in urlopen

    body=body, headers=headers)

  File "/Users/travis/miniconda3/lib/python3.5/site-packages/requests/packages/urllib3/connectionpool.py", line 378, in _make_request

    httplib_response = conn.getresponse()

  File "/Users/travis/miniconda3/lib/python3.5/http/client.py", line 1197, in getresponse

    response.begin()

  File "/Users/travis/miniconda3/lib/python3.5/http/client.py", line 297, in begin

    version, status, reason = self._read_status()

  File "/Users/travis/miniconda3/lib/python3.5/http/client.py", line 266, in _read_status

    raise RemoteDisconnected("Remote end closed connection without"

requests.packages.urllib3.exceptions.ProtocolError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response',))



During handling of the above exception, another exception occurred:



  File "/Users/travis/miniconda3/lib/python3.5/site-packages/binstar_client/commands/upload.py", line 214, in main

    callback=upload_print_callback(args))

  File "/Users/travis/miniconda3/lib/python3.5/site-packages/binstar_client/__init__.py", line 494, in upload

    s3res = requests.post(s3url, data=data_stream, verify=self.session.verify, timeout=10 * 60 * 60, headers=headers)

  File "/Users/travis/miniconda3/lib/python3.5/site-packages/requests/api.py", line 107, in post

    return request('post', url, data=data, json=json, **kwargs)

  File "/Users/travis/miniconda3/lib/python3.5/site-packages/requests/api.py", line 53, in request

    return session.request(method=method, url=url, **kwargs)

  File "/Users/travis/miniconda3/lib/python3.5/site-packages/requests/sessions.py", line 468, in request

    resp = self.send(prep, **send_kwargs)

  File "/Users/travis/miniconda3/lib/python3.5/site-packages/requests/sessions.py", line 576, in send

    r = adapter.send(request, **kwargs)

  File "/Users/travis/miniconda3/lib/python3.5/site-packages/requests/adapters.py", line 426, in send

    raise ConnectionError(err, request=request)

requests.exceptions.ConnectionError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response',))



During handling of the above exception, another exception occurred:



  File "/Users/travis/miniconda3/bin/anaconda", line 6, in <module>

    sys.exit(binstar_client.scripts.cli.main())

  File "/Users/travis/miniconda3/lib/python3.5/site-packages/binstar_client/scripts/cli.py", line 87, in main

    description=__doc__, version=version)

  File "/Users/travis/miniconda3/lib/python3.5/site-packages/binstar_client/scripts/cli.py", line 66, in binstar_main

    return args.main(args)

  File "/Users/travis/miniconda3/lib/python3.5/site-packages/binstar_client/commands/upload.py", line 219, in main

    except requests_ext.OpenSslError:

TypeError: catching classes that do not inherit from BaseException is not allowed

  File "./ci_support/upload_or_check_non_existence.py", line 124, in <module>

    main()

  File "./ci_support/upload_or_check_non_existence.py", line 108, in main

    upload(cli, meta, owner, channel)

  File "./ci_support/upload_or_check_non_existence.py", line 55, in upload

    env=os.environ)

  File "/Users/travis/miniconda3/lib/python3.5/subprocess.py", line 581, in check_call

    raise CalledProcessError(retcode, cmd)

subprocess.CalledProcessError: Command '['anaconda', '--quiet', '-t', 'binstar.token', 'upload', '/Users/travis/miniconda3/conda-bld/osx-64/scipy-0.17.1-np110py35_blas_openblas_201.tar.bz2', '--user=conda-forge', '--channel=main']' returned non-zero exit status 1



travis_time:end:0e2e2414:start=1468271674000000000,finish=1468271720000000000,duration=46000000000
[0Ktravis_fold:end:after_success
[0K

Done. Your build exited with 0.
```

jakirkham:
@pelson, just wanted you to be aware of the uploading issue that occurred. Though it can be investigated later.

jakirkham:
After closer inspection, it seems this only affected the Python 3.5, NumPy 1.10 build. So, that one has been restarted.

PR #4, opened 2016-07-11
Unpin conda-build on Travis CI [ci skip]

jakirkham:
Not sure if we are going to need this or not. In some cases we have had bad symlinks getting packaged when using activation scripts (like those in the `toolchain`) like as shown in this issue ( https://github.com/conda-forge/hdf5-feedstock/issues/23 ). Though we haven't seen this problem crop up recently and are not totally sure why. There may have been a change in `conda` that has made this unnecessary, but we have been unable to confirm. So, am just readying this in case it is needed.

xref: https://github.com/conda-forge/scipy-feedstock/pull/2

cc @pelson @rgommers @gillins

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ install:
 
       conda config --set show_channel_urls true
       conda update --yes conda
-      conda install --yes conda-build=1.20.0 jinja2 anaconda-client
+      conda install --yes conda-build jinja2 anaconda-client
       conda config --add channels conda-forge
       
 


jakirkham:
I have been inspecting the binaries produced for OS X and they don't seem to need this unpinning. While I still have no idea why this isn't needed now, I am glad to see everything is working smoothly. Will drop the build number bump, but I think we should put this in just in case.

rgommers:
Seems like a good idea. I'd expect that we want to use the latest ``conda-build`` version everywhere for consistency, unless a specific build issue needs a workaround.

going to leave it open for a bit in case someone with more knowledge of the history wants to comment, but +1 from me.

jakirkham:
> Seems like a good idea. I'd expect that we want to use the latest ``conda-build`` version everywhere for consistency...

There are some tricky points (e.g. 2.0.0), which we will want to embrace cautiously. Though generally I agree that we should try to use the newest `conda-build`. We are moving to handling these pinning workarounds via [`conda-forge-build-setup`]( https://github.com/conda-forge/conda-forge-build-setup-feedstock ) so that we can control this more effectively in one location.

> ...unless a specific build issue needs a workaround.

This particular pinning is due to a relatively ancient (in conda-forge terms 😉) problem ( https://github.com/conda/conda-build/issues/889 ), which has since been fixed. We have just become reluctant to update all pinnings in feedstocks due to the relatively high churn and the effects from that (e.g. CI backlog).

> going to leave it open for a bit in case someone with more knowledge of the history wants to comment, but +1 from me.

This particular pinning caused us issues as it did not respect behavior in newer versions of `conda` and `conda-env`. So, it resulted in packaging some pieces of the environment, which it shouldn't. We have noticed this mainly with packages that use activation scripts (e.g. the `toolchain`). The results were pretty devastating as it resulted in packages failing to install. Here is the first case of that ( https://github.com/conda-forge/hdf5-feedstock/issues/23 ). Fortunately, the fix was easy ( https://github.com/conda-forge/hdf5-feedstock/pull/25 ) and we demonstrated it worked in short order. However, this problem seems to have mysteriously vanished. We think there was some change in `conda`/`conda-env` maybe something subtle like a timing issue, but never narrowed it down. @gillins and I have taken no chances and simply unpin this anyways for good measure.

pelson:
@jakirkham - did you consider re-rendering with conda-smithy?

jakirkham:
Sadly that won't fix the problem if we are still pinned in the `conda-forge-build-setup` package. Or did that PR get merged and I missed it?

jakirkham:
Though I did re-render with `conda-smithy` in another case with `scipy`. It dropped the versioning of `python` and `numpy`. When I went back to repeat it, I was unable to. I really wish I could tell you how it happened. I may try to reproduce it if you haven't already narrowed down the problem.

jakirkham:
Is there any pressing reason not to merge this? AFAICT this is fine to go.

jakirkham:
I've added `[ci skip]` to the title so that hopefully no CI is run when this is merged. Given how long it takes to run the CI for `scipy` (particularly on Linux which runs it all in a single build), I think we should avoid doing it if it is not productive. This case of merging seems non-productive for CI as we have already determined here that it works and we are not trying to produce binaries from it.

jakirkham:
@pelson, this is the PR ( https://github.com/conda-forge/conda-forge-build-setup-feedstock/pull/14 ) I was referring to.

PR #6, opened 2016-07-26
Rebuild with parallel OpenBLAS

jakirkham:
Bump build number and rebuild with the parallel OpenBLAS.

cc @ocefpaf

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -15,6 +15,11 @@ unset LDFLAGS
 source activate "${CONDA_DEFAULT_ENV}"
 
 
+# Use OpenBLAS with 1 thread only as it seems to be using too many
+# on the CIs apparently.
+export OPENBLAS_NUM_THREADS=1
+
+
 export LIBRARY_PATH="${PREFIX}/lib"
 export C_INCLUDE_PATH="${PREFIX}/include"
 export CPLUS_INCLUDE_PATH="${PREFIX}/include"
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -13,7 +13,7 @@ source:
   md5: d70e7f533622ab705bc016dac328d93e
 
 build:
-  number: 200
+  number: 201
   # We lack openblas on Windows, and therefore can't build scipy there either currently.
   skip: true  # [win]
   features:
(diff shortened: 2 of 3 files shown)


jakirkham:
This is running into issues on CIs because it is using too many threads I believe. This [log]( https://travis-ci.org/conda-forge/scipy-feedstock/jobs/147511310#L16713 ) is what I'm using as evidence. Am trying to set the number of threads to 1 in `build.sh` to workaround this issue.

recipe/meta.yaml
```diff
@@ -40,13 +40,6 @@ test:
   requires:
     - nose
     - mpmath
-  imports:
-    - scipy
-    - scipy.integrate.vode
-    - scipy.linalg._fblas
-    - scipy.linalg._flapack
-    - scipy.sparse.linalg.eigen.arpack._arpack
-    - scipy.sparse.linalg.isolve._iterative
```
jakirkham: Dropped as they are all imported in `run_test.py` (checked each one). Also importing them before setting `OPENBLAS_NUM_THREADS` may ignore that setting change.

recipe/run_test.py
```diff
@@ -1,4 +1,9 @@
 import sys
+import os
+
+# Use OpenBLAS with 1 thread only as it seems to be using too many
+# on the CIs apparently.
+os.environ["OPENBLAS_NUM_THREADS"] = "1"
```
jakirkham: Sets `OPENBLAS_NUM_THREADS` to `1`. This should stop the segmentation fault on Travis CI.

recipe/build.sh
```diff
@@ -15,6 +15,11 @@ unset LDFLAGS
 source activate "${CONDA_DEFAULT_ENV}"
 
 
+# Use OpenBLAS with 1 thread only as it seems to be using too many
+# on the CIs apparently.
+export OPENBLAS_NUM_THREADS=1
```
jakirkham: Didn't need it in the build before, but just setting it for good measure. Shouldn't have any effect on how OpenBLAS is used at run-time.

jakirkham:
Yay! Travis builds appear to be working so far. 🎉

jakirkham:
Alright this is passing.

jakirkham:
Ack! Just barely ran out of build time on [CircleCI]( https://circleci.com/gh/conda-forge/scipy-feedstock/41 ). Have restarted the build and am hoping it is a little faster this time. However, it looks like the build times on that platform are getting quite close to the limit anyways. See the 0.18.0 [build]( https://circleci.com/gh/conda-forge/scipy-feedstock/37 ) on CircleCI last night took nearly 2hrs (the build limit). We might need to figure out a new way to do this if we need all of these builds.

jakirkham:
Opened this issue ( https://github.com/conda-forge/scipy-feedstock/issues/8 ) about being close to the CircleCI limit.

PR #9, opened 2016-08-06
Rebuild with parallel OpenBLAS

jakirkham:
Bump build number and rebuild with the parallel OpenBLAS. Basically the same as PR ( https://github.com/conda-forge/scipy-feedstock/pull/6 ) for `0.17.x`. To see a use case for this rebuild, please see PRs ( https://github.com/conda-forge/pytmatrix-feedstock/pull/4 ), ( https://github.com/conda-forge/wradlib-feedstock/pull/16 ) where it appears `scipy` version `0.18.0` is not supported yet.

cc @ocefpaf @kmuehlbauer

### DIFF
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -15,6 +15,11 @@ unset LDFLAGS
 source activate "${CONDA_DEFAULT_ENV}"
 
 
+# Use OpenBLAS with 1 thread only as it seems to be using too many
+# on the CIs apparently.
+export OPENBLAS_NUM_THREADS=1
+
+
 export LIBRARY_PATH="${PREFIX}/lib"
 export C_INCLUDE_PATH="${PREFIX}/include"
 export CPLUS_INCLUDE_PATH="${PREFIX}/include"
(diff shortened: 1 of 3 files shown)


jakirkham:
So the build on CircleCI keeps timing out. I keep restarting it in the hopes that it may complete within the time limit, but I think we need to do something to cutdown the build time. An easy step along these lines would be to include the conda-forge channel in the docker image via PR ( https://github.com/conda-forge/docker-images/pull/24 ). This should cutdown setup time. Though ultimately we will want some sort of build matrix for CircleCI as described in issue ( https://github.com/conda-forge/conda-smithy/issues/203 ) to cutdown the build time.

jakirkham:
Rebuilding CircleCI with the new Docker container.

jakirkham:
So this is still timing out. I don't see a way forward with the current configuration and get matrix builds working will take more time than I can allot to this currently. I'd like to propose we split the SciPy version branches further based on supported NumPy version (e.g. `v0.17_np1.10` or similar and so on). This should be easy to do and will get this up and working quickly. Please let me know your thoughts.

cc @pelson @ocefpaf

jakirkham:
Closing as this PR has no way to proceed currently due to CI limits. Proposing we break up the NumPy builds as seen in PRs ( https://github.com/conda-forge/scipy-feedstock/pull/10 ) and ( https://github.com/conda-forge/scipy-feedstock/pull/11 ).

PR #10, opened 2016-08-11
Config SciPy v0.17 build with NumPy 1.11

jakirkham:
As it seems to be too much to build 2 NumPy versions and 3 Python versions on CircleCI, I'm proposing we break them up into 2 branches that build with 1 NumPy each. This should be well within the time limit. Think of this as a poor man's build matrix. In the long run, we will want to have some sort of build matrix support via CircleCI's parallelism feature as explained in issue ( https://github.com/conda-forge/conda-smithy/issues/203 ). However, after spending some time on it, it is clear to me this is going to take a bit more work. Unfortunately, I don't have the time to dedicate to that problem at present. So, I'm punting on it by breaking these up in the hopes of solving downstream package build issues seen in PRs ( https://github.com/conda-forge/pytmatrix-feedstock/pull/4 ) and ( https://github.com/conda-forge/wradlib-feedstock/pull/16 ) (potentially others). This includes the build number bump to force a rebuild around the parallel OpenBLAS. The NumPy 1.10 build can be seen in PR ( https://github.com/conda-forge/scipy-feedstock/pull/11 ).

cc @pelson @ocefpaf @kmuehlbauer

PR #17, opened 2016-08-31
Config SciPy v0.18 build with NumPy 1.11

jakirkham:
This extends the same strategy used with SciPy v0.17 to v0.18 as was proposed for the NumPy 1.11 build in PR ( https://github.com/conda-forge/scipy-feedstock/pull/10 ). Also absorbs PR ( https://github.com/conda-forge/scipy-feedstock/pull/12 ) while we are at it. The NumPy 1.10 build is in PR ( https://github.com/conda-forge/scipy-feedstock/pull/16 ).

Additionally I propose that `master` stays up-to-date with this branch. There are many cases where a `master` branch is required. As a result, it is merely a matter of choosing where it is. Having it follow the latest version of SciPy with the latest version of NumPy makes a lot of sense.

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -13,7 +13,7 @@ source:
   md5: d70e7f533622ab705bc016dac328d93e
 
 build:
-  number: 201
+  number: 202
   # We lack openblas on Windows, and therefore can't build scipy there either currently.
   skip: true  # [win]
   features:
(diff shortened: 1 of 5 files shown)


recipe/meta.yaml
```diff
@@ -24,6 +24,7 @@ requirements:
     - toolchain
     - gcc          # [osx]
     - python
+    - setuptools
```
jakirkham: Discovered this was using `setuptools` after re-rendering. See this [build]( https://circleci.com/gh/conda-forge/scipy-feedstock/78 ) as evidence. As a result, we have added it here and the need arguments above.

jakirkham:
Passing. 💚

jakirkham:
Merged as this is a change we already accepted in principle with v0.17. Otherwise an uniteresting change as it is just a re-rendering and slight tweak to the OpenBLAS pinning.

jakirkham:
Have deleted `v0.18_np1.11` as `master` is on the exact same commit. As we need to keep `master` and it is best if it is on the latest version, it seems that `v0.18_np1.11` is redundant. No need to have more pinning or re-rendering PRs than necessary. 😄

PR #42, opened 2017-01-05
py36

ocefpaf:


ci_support/run_docker_build.sh
```diff
@@ -70,4 +70,11 @@ yum install -y devtoolset-2-gcc-gfortran
     set +x
     conda build /recipe_root --quiet || exit 1
     upload_or_check_non_existence /recipe_root conda-forge --channel=main || exit 1
+
+    set -x
+    export CONDA_NPY=110
```
rgommers: Are we really doing numpy 1.10 for Python 3.6? Numpy itself doesn't support that (although it may work with maybe some test failures). The first wheels for Python 3.6 are when numpy 1.12.0 comes out I think.

jakirkham: I don't think we plan to maintain NumPy 1.10 for much longer. Though I don't believe it has been formally discussed, I expect we will drop it after NumPy 1.12 lands.

As for Python 3.6, in particular, we do run the test suite in both builds of NumPy and SciPy. So if it passed, it should work, right? Were there many notable changes needed to NumPy and SciPy for Python 3.6 support?

rgommers: Just a whole bunch of warnings were fixed, no breakage: 
https://github.com/numpy/numpy/pull/8391 
https://github.com/numpy/numpy/pull/8359
https://github.com/numpy/numpy/pull/8178

ocefpaf: @rgommers I agree and even try to drop it already. But some people think we should and are blocking my PRs if I do not add it :unamused:

PR #43, opened 2017-01-07
Use SHLIB_EXT

jschueller:


recipe/build.sh
```diff
@@ -30,12 +30,8 @@ if [[ `uname` == 'Darwin' ]]; then
     # `gfortran` and cause problems.
     #
     # https://github.com/conda-forge/toolchain-feedstock/pull/8
-    DYLIB_EXT=dylib
     export CFLAGS="${CFLAGS} -stdlib=libc++ -lc++"
     export LDFLAGS="-headerpad_max_install_names -undefined dynamic_lookup -bundle -Wl,-search_paths_first -lc++"
-else
-    DYLIB_EXT=so
-    unset LDFLAGS
```
jakirkham: We cannot drop the `unset` here. Could you please add it back?

jschueller: it's already there at the beginning

jakirkham: Yeah, the beginning one should probably be removed. The point is this can't be set to anything on Linux (even empty string) or SciPy won't configure it correctly. However, the toolchain does set this. So we override it here.

jakirkham:
Thanks @jschueller . Definitely a cleaner approach. 👍

PR #44, opened 2017-01-17
Re-render with auto...

ocefpaf:
... to get NumPy 1.12

PR #45, opened 2017-01-17
Remove the numpy skip

ocefpaf:
Now that my test in #44 worked we can remove the skip.


jakirkham:
Seems this is consistently failing on `master` now. My best guess is that we weren't actually running with NumPy 1.12 because this selector blocked. So I guess we need to change this selector to 1.12 and add a new branch for 1.11.

PR #47, opened 2017-01-25
Build only np111 for Linux on this branch

ocefpaf:
See #48


ocefpaf:
No need to wait for Travis-CI as this affects only Linux.

PR #48, opened 2017-01-25
Build only np112 for Linux on master

ocefpaf:
See https://github.com/conda-forge/scipy-feedstock/pull/47


ocefpaf:
No need to wait for Travis-CI as this affects only Linux.

Issue #59, opened 2017-05-02
libgcc_s.dylib linking problem


mingwandroid:
Incorrectly reported here: https://github.com/ContinuumIO/anaconda-issues/issues/1618

mingwandroid:
```
find $CONDA_PREFIX/lib/python*/site-packages/scipy -name "*.so" | xargs otool -L | grep libgcc_s
	/Users/ray/mc-x64-3.5/conda-bld/gcc-4.8_1477649012852/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/libgcc_s.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/Users/ray/mc-x64-3.5/conda-bld/gcc-4.8_1477649012852/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/libgcc_s.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/Users/ray/mc-x64-3.5/conda-bld/gcc-4.8_1477649012852/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/libgcc_s.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/Users/ray/mc-x64-3.5/conda-bld/gcc-4.8_1477649012852/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/libgcc_s.1.dylib (compatibility version 1.0.0, current version 1.0.0)
	/Users/ray/mc-x64-3.5/conda-bld/gcc-4.8_1477649012852/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/libgcc_s.1.dylib (compatibility version 1.0.0, current version 1.0.0)
[...]

PR #60, opened 2017-05-02
bump build to build against newest macOS GCC

mingwandroid:


jakirkham:
If it helps, then that's fine, but I don't follow why it is needed.

xref: https://github.com/conda-forge/scipy-feedstock/issues/59#issuecomment-298675570

mingwandroid:
I have no merge rights to this, please go ahead. This rebuild is necessary (you could try using the current `scipy` package on macOS if you need to see it with your own eyes).

jakirkham:
So I did just try the current SciPy package with Python 2.7, 3.5, and 3.6 using a totally fresh environment. Was unable to reproduce the issue shown in the bug report. Is there something else I need to be doing?

xref: https://github.com/ContinuumIO/anaconda-issues/issues/1618

mingwandroid:
I can't get it to repeat either actually, but the current dylibs are without doubt, broken.

The following command:

```
find $CONDA_PREFIX/lib/python*/site-packages/scipy -name "*.so" -exec bash -c " echo {} ; otool -l {} | grep \" name \"; echo \"\" " \;
```

.. gives a list of the load commands embedded in the scipy .so files. As you can see, many are contain hardcoded absolute paths to a folder on my old build machine. IMHO, the issue is not exhibiting because our macOS versions are loading `libgcc_s.dylib` transitively via `@rpath/libgfortran.3.dylib` before encountering this broken load command and it doesn't try to load the same one twice.

I really think you should fix this. If you were to run a tool to check the sanity of your dylibs these missing load command targets would show up because they are broken.

I'm out of time to spend on this, please just merge it.

satra:
just encountered this issue on a straight up scipy install on osx from conda-forge.

```
 <module>
    from scipy import linalg
../miniconda3/envs/testcf/lib/python3.6/site-packages/scipy/linalg/__init__.py:175: in <module>
    from .misc import *
../miniconda3/envs/testcf/lib/python3.6/site-packages/scipy/linalg/misc.py:5: in <module>
    from .blas import get_blas_funcs
../miniconda3/envs/testcf/lib/python3.6/site-packages/scipy/linalg/blas.py:155: in <module>
    from scipy.linalg import _fblas
E   ImportError: dlopen(/software/miniconda3/envs/testcf/lib/python3.6/site-packages/scipy/linalg/_fblas.cpython-36m-darwin.so, 2): Library not loaded: /usr/local/lib/libgcc_s.1.dylib
E     Referenced from: /software/miniconda3/envs/testcf/lib/python3.6/site-packages/scipy/linalg/_fblas.cpython-36m-darwin.so
E     Reason: image not found
```

jakirkham:
Sorry to get stuck in minutiae @mingwandroid. I'm more than happy to merge this. Did we need this for anything else? Guessing not NumPy. Maybe scikit-learn?

mingwandroid:
I don't think anything else needs a rebuild. I'll try to check when I get a chance though.

jakirkham:
Ok, thanks. Also do we need this for the NumPy 1.11 branch as well or did this only occur with the NumPy 1.12 build?

PR #75, opened 2017-09-28
Build test for scipy 1.0.0

ocefpaf:
~~Do not merge! Just testing...~~

Based on the latest 1.0.0 release now :tada:

PR #76, opened 2017-10-25
Update to 1.0.0

rgommers:


ocefpaf:
Thanks @rgommers!

I have #75 with more changes that will probably be needed. Feel free to close that one and port the changes here.

rgommers:
Ah nice. Missed your PR, sorry. Looks like the ``pytest`` change is definitely needed, and the OpenBLAS version bump seems like a good idea too. Is the rest needed, or is that orthogonal?

ocefpaf:
The rerendering can be done automatically with:

Some of the tests will need the modifications because `.wasSuccessful()` and `scipy.linalg._calc_lwork` seems to be removed in `1.0.0`.

rgommers:
Does it make more sense to close this one and merge gh-75?

ocefpaf:
> Does it make more sense to close this one and merge gh-75?

Up to you I am OK either way :smile:

rgommers:
Your PR was first and looks good, so let's go with that.

PR #87, opened 2018-12-05
[WIP] remove blas_openblas feature

minrk:
it is being hotfixed out of the blas metapackage so we don't need it anymore (and indeed it *must* be removed to be installable), but we couldn't remove it until blas was hotfixed to remove track_features, which it is now

same as https://github.com/conda-forge/numpy-feedstock/pull/121

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -10,10 +10,8 @@ source:
   sha256: 0291a2254d9a020af9b49ca648e8b41e1366be9a8a40849be9102626b2022422
 
 build:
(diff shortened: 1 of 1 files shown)


minrk:
Maybe cancelling this since the hotfix is not being applied to the gcc7 channel, which means the feature *is* needed for the gcc7 channel.

https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/3 reverts the hotfix for blas_openblas, making this feature required once again for any package that has been built in the past with the feature.

PR #93, opened 2019-02-18
Update to 1.2.1

jschueller:
I understood what happened with the failling test in https://github.com/scipy/scipy/issues/9777

They fixed it with:
https://github.com/tylerjereddy/scipy/commit/ef1659a0a9b654e1c9c2ab72993b5c689f733532

But we build with np<1.14 and run the tests with np1.14, so that test still needs to be disabled.


ocefpaf:
> But we build with np<1.14 and run the tests with np1.14, so that test still needs to be disabled.

OK. Let's get this merged to get SciPy out. Thanks for fixing this @jschueller!

PR #114, opened 2019-11-03
Rebuild python3801

chrisburr:
Closes #113 

So far just adding the patch that was needed for numpy: https://github.com/conda-forge/numpy-feedstock/pull/174/commits/49a2950d71d3e36155e19e70ebecf713ae9db64f


isuruf:
We should fix https://github.com/conda-forge/python-feedstock/issues/281 instead of patching packages like this.

chrisburr:
> We should fix [conda-forge/python-feedstock#281](https://github.com/conda-forge/python-feedstock/issues/281) instead of patching packages like this.

I've rebase 0b4bff4 out of the history now that it is fixed.

chrisburr:
The 3.8 deadlock is probably https://github.com/scipy/scipy/issues/10927

hmaarrfk:
it seems that we should just wait for the new release

beckermr:
If someone can point me to the exact patch we need, I'm happy to port it over. There are 103 commits to the maintenance branch and picking them apart is not easy.

beckermr:
OK. I've pushed a patch of everything on the `maintenance/1.3.x` branch minus a few changes in their CI. Let's see what gives.

beckermr:
@ocefpaf for viz

beckermr:
Looks like aarch and ppc64le might still be broken, but linux passed on 3.8!

beckermr:
Looks like osx on python 3.8 is still hanging.

ocefpaf:
Thanks @chrisburr and @beckermr for putting this together. Let's focus our efforts on #113 now that we have a release.

chrisburr:
I agree but shouldn’t this be left open until #113 is merged to prevent the bot from continuing?

ocefpaf:
I guess that only #113 is in the bot's database.

chrisburr:
Yes you’re right. I forgot I made a new PR so I could push changes.

PR #155, opened 2021-01-11
Manual pypy37 migration

h-vetinari:
There's some weird cycles in the dependency graph that make the migrator think that scipy depends on pandas and psycopg2, among others, see https://github.com/regro/cf-scripts/issues/1331

I verified that all dependencies have been built (hopefully not overlooking anything), so raising this to break some of these cycles.

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -16,6 +16,10 @@ jobs:
         CONFIG: linux_64_numpy1.16python3.8.____cpython
         UPLOAD_PACKAGES: 'True'
         DOCKER_IMAGE: quay.io/condaforge/linux-anvil-comp7
+      linux_64_numpy1.19python3.7.____73_pypy:
+        CONFIG: linux_64_numpy1.19python3.7.____73_pypy
+        UPLOAD_PACKAGES: 'True'
+        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-comp7
       linux_64_numpy1.19python3.9.____cpython:
         CONFIG: linux_64_numpy1.19python3.9.____cpython
         UPLOAD_PACKAGES: 'True'
--- a/.azure-pipelines/azure-pipelines-osx.yml
+++ b/.azure-pipelines/azure-pipelines-osx.yml
@@ -14,6 +14,9 @@ jobs:
       osx_64_numpy1.16python3.8.____cpython:
         CONFIG: osx_64_numpy1.16python3.8.____cpython
         UPLOAD_PACKAGES: 'True'
+      osx_64_numpy1.19python3.7.____73_pypy:
+        CONFIG: osx_64_numpy1.19python3.7.____73_pypy
+        UPLOAD_PACKAGES: 'True'
       osx_64_numpy1.19python3.9.____cpython:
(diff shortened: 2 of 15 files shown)


beckermr:
Thank you @h-vetinari!

h-vetinari:
The last drone job timed out - I've seen this pattern a lot in the blas-variant PRs: 2-3 drone jobs go through in 25-30min, the ones beyond that get less resources(?) and then ultimately fail at the 60min timeout (somewhere between finishing the build and the middle of the test suite). Let's see if following https://github.com/scipy/scipy/blob/v1.6.0/setup.py#L234-L237 has any effect.

beckermr:
There are two different machines that drone runs jobs on. One is faster than the other I think? @isuruf has details

h-vetinari:
> @beckermr: There are two different machines that drone runs jobs on. One is faster than the other I think? @isuruf has details

How much control do we have over this process? If the timeout were 90min or so, the build would very likely pass even on the slow machines (instead of retrying it all the time and spending much more CI time in total).

beckermr:
> How much control do we have over this process? If the timeout were 90min or so, the build would very likely pass even on the slow machines (instead of retrying it all the time and spending much more CI time in total).

None. Yay!

h-vetinari:
> None. Yay!

Then I suggest to merge as is and manually restart the drone build once or twice until all variants get built eventually.

beckermr:
OK. Let's let it get as far as it can to make sure the tests pass. Then we can merge.

h-vetinari:
> 
> 
> OK. Let's let it get as far as it can to make sure the tests pass. Then we can merge.

I just removed the patch:
```diff
commit 07dd65932da4c773990bb6480503724ba78b74a4
Author: H. Vetinari <h.vetinari@gmx.com>
Date:   Mon Jan 11 20:29:17 2021 +0100

    try upping build parallelism

diff --git a/recipe/build.sh b/recipe/build.sh
index 2ebd6d5d5..9cf1c5f4a 100644
--- a/recipe/build.sh
+++ b/recipe/build.sh
@@ -4,6 +4,7 @@
 # can have a G77 ABI (currently only MKL)
 export SCIPY_USE_G77_ABI_WRAPPER=1

+set NPY_NUM_BUILD_JOBS=${CPU_COUNT}
 if [[ "$python_impl" == "pypy" && "$target_platform" == "linux-ppc64le" ]]; then
     $PYTHON setup.py install --single-version-externally-managed --record=record.txt
 else
```

The pipeline with this exact state had already run through before, with everything but one aarch-build passing. Proof: compare first commit in "merge xxx into yyy" [before](https://github.com/conda-forge/scipy-feedstock/commit/c3a9ad97232334d5f358cb50c95187278aa962aa) (clickable from [here](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=261920&view=results)), and [after](https://github.com/conda-forge/scipy-feedstock/commit/c9443ee0c434fb0a58559133d1af6615cca98ab6) (clickable from [here](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=262139&view=results))

I say this because aarch builds will almost certainly fail again, and so the automerge label will be useless. Which of the 4 aarch jobs fails is pretty random, so IMO this can be handled _after_ merging by restarting until the missing job also runs through on the faster machine.

PR #165, opened 2021-03-27
POC: Build windows with vs2019

h-vetinari:
Checklist

In https://github.com/scipy/scipy/pull/13713, the discussion came up of using the universal C runtime (since vs2015) to uncouple the compiler version that scipy uses for windows from the one used to build CPython itself.

@rgommers is wary of doing this without knowing what subtle bugs this will incur, and one option that came up was demonstrating this in conda-forge. AFAICT, conda-forge _already_ relies on the UCRT (cf. https://github.com/conda-forge/vc-feedstock/blob/master/recipe/meta.yaml), because some packages are selectively being built with vs2019 already, while CPython obviously isn't.

This PR is to demonstrate just that; IMO it could be merged, but that might a bit more discussion. I'll also open a PR for the scipy CI itself.

### DIFF
--- a/.azure-pipelines/azure-pipelines-win.yml
+++ b/.azure-pipelines/azure-pipelines-win.yml
@@ -5,7 +5,7 @@
 jobs:
 - job: win
   pool:
-    vmImage: vs2017-win2016
+    vmImage: windows-2019
   strategy:
     matrix:
       win_64_numpy1.17python3.7.____cpython:
--- a/conda-forge.yml
+++ b/conda-forge.yml
@@ -4,5 +4,7 @@ provider: {linux_aarch64: default, linux_ppc64le: default}
 test_on_native_only: true
 azure:
   settings_win:
+    pool:
+      vmImage: windows-2019
     variables:
       CONDA_BLD_PATH: "C:\\\\bld\\\\"
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -9,7 +9,7 @@ source:
   sha256: adfc9a8457ea16c08acd2cd1e54f4826ca8e7345060d1cc4c879b6f26c400c66
 
 build:
-  number: 0
+  number: 1
   skip: true  # [py2k or py<=36]
 
 requirements:


isuruf:
@h-vetinari, @rgommers, while it's okay that conda packages be built with vs2019, wheels should be built with the same compiler version as python to avoid users having to download a newer Visual C runtime.

h-vetinari:
@isuruf

That's exactly the point of the universal runtime though (notwithstanding how it's implemented in conda-forge) - the runtime shouldn't have to change in principle. See e.g. [this explanation](https://discuss.python.org/t/toolchain-upgrade-on-windows/6377/4) by Steve Dower, which I've tried to capture in a few more words in https://github.com/scipy/scipy/pull/13713 (where Steve was tagged, but didn't comment).

isuruf:
It does change with C++ and scipy has C++ files in it.

isuruf:
Also, if there are static libraries like `npymath.lib`, then downstream packages have to use the same version or newer version of the linker when linking that library because even though VS2015/2017/2019 are binary compatible, you need to use the newest among the linking object/libraries.

Quoting https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=msvc-160,

> You can mix binaries built by different versions of the v140, v141, and v142 toolsets. However, you must link by using a toolset at least as recent as the most recent binary in your app. Here's an example: you can link an app compiled using any 2017 toolset (v141, versions 15.0 through 15.9) to a static library compiled using, say, Visual Studio 2019 version 16.2 (v142), if they're linked using a version 16.2 or later toolset. You can link a version 16.2 library to a version 16.4 app as long as you use a 16.4 or later toolset.

h-vetinari:
Thanks for the infos @isuruf, much appreciated, because I don't understand all the ins & outs.

I had asked about the scipy case specifically (see below), but it seems we're back to square one scipy remaining tied to the compiler version used for CPython, which brings additional relevance back to https://bugs.python.org/issue42380.

> > @h-vetinari: So IIUC [...] things should work more or less out of the box when using a newer MSVC to compile e.g. scipy for python 3.6?
>
> Steve Dower: Yep, and if they don’t, it’s a compiler/runtime bug that I can report to the team at Microsoft (not that a lot can be done about some things).
>
> The actual compiler version used for CPython increases whenever I remember to update my build machine, but it’ll stay on v14 for anything that’s already released. Last I heard the MSVC team have no plans to up their version, so the next one should be a while away.

h-vetinari:
Added a new comment on this topic on discuss.python.org, hoping that Steve will chime in.

isuruf:
Note that I didn't mean to say that UCRT needs to be updated. I mean Visual Studio C++ runtime.

h-vetinari:
OK, thanks. Feel free to jump in on the discuss topic - I'm clearly a lossy middleman. 🙃

h-vetinari:
Upstream moved to vs2019 recently, so this can be closed.

PR #168, opened 2021-05-09
Cleanup run_test.py & other lint

h-vetinari:
The automatically-run `run_test.py` didn't get many changes in recent years; the last changes were in #124, and https://github.com/conda-forge/scipy-feedstock/commit/6b0ccab54832b9084a1d7d94ff07dec631ff8500 made a bunch of stuff obsolete.

I'd also say that the imports should move out into the recipe section (not least in light of #160), where the import-tests should be with the main scipy package, and the `run_test.py` run for `scipy-tests` only.

PR #146, opened 2020-10-31
Windows with gfortran

isuruf:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

Fixes #80
Fixes #119
Fixes #136
Fixes #139

PR #140, opened 2020-10-14
[WIP] M2sys for windows -- it seems that scipy does this upstream

hmaarrfk:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

Issue #80, opened 2018-05-02
SciPy builds for Windows


FSund:
~~Is it possible to add OpenBLAS builds of scipy for Windows?~~ As noted by @tadeu there is no `scipy` package for Windows in `conda-forge`, so the proper request to make here the addition of (OpenBLAS-enabled) Windows builds of SciPy to `conda-forge`. This would help a lot with the size of executables made from Python programs on Windows.
 
The MKL libraries installed by Anaconda on Windows when installing scipy are very large, so if making an executable (using PyInstaller or similar), even a simple `import scipy; print("Hello, world!")` ends up at around 200 MB.

A similar issue for numpy can be found here: https://github.com/conda-forge/numpy-feedstock/issues/84

tadeu:
Try using the `nomkl` feature (i.e., install the `nomkl` package together), but note that this do not have to do with `conda-forge` because `conda-forge` do not have packages that depend on `MKL` (you're probably using `scipy` from the `defaults` Anaconda channel).  In fact `conda-forge` currently do not even have `scipy` packages for Windows :(

FSund:
You are both completely correct. There is no `nomkl` package for Windows, and `conda-forge` doesn't have  a `scipy` package for Windows. I thought I installed `scipy` from `conda-forge`, but now see that the package comes from the regular channel.

So **SciPy builds for Windows** is probably a much better title for this issue.

FSund:
I thought I'd try my hand at getting a Windows version up, but I think I'm in way over my head. Anyway, here's what I have learned so far:
* There seems to be a [OpenBLAS that has recipies for Windows](https://github.com/conda-forge/openblas-feedstock), contrary to what [this comment](https://github.com/conda-forge/scipy-feedstock/blob/master/recipe/meta.yaml#L14) says
  * I might be wrong about this, I have never read a yaml-file or any kind of feedstock before
* The `conda-forge` version of [numpy](https://github.com/conda-forge/numpy-feedstock) uses OpenBLAS on Windows. 

This indicates to me that it shouldn't be *too* hard to get SciPy+OpenBLAS working on Windows. I might be horribly wrong here though. Has there been any attempts at this?

[...]

PR #169, opened 2021-06-06
scipy v1.7.0

h-vetinari:
Also picking up some of the good bits from #168

TODOs:

- [x] build new dependency (pythran) on all platforms
- [x] figure out how to deal with new submodule
  - [x] ~follow-up: switch to sdists from GH (as soon as 1.7.0rc2 appears)~
- [x] fix windows builds
- [ ] fix PyPy builds
- [ ] fix PPC builds


h-vetinari:
Interesting new error...
```
      File "scipy/_lib/setup.py", line 9, in check_boost_submodule
        raise RuntimeError("Missing the `boost` submodule! Run `git submodule "
    RuntimeError: Missing the `boost` submodule! Run `git submodule update --init` to fix this.
```

This looks like an additional dependency: https://github.com/scipy/scipy/commit/bbc193684f763bbe532da2a6a3fbdd91802707f2, which now expectes a git-submodule in [`scipy/_lib`](https://github.com/scipy/scipy/tree/v1.7.0rc1/scipy/_lib), which runs - of course - into the eternal pain that is dear-github/dear-github#214.

Let's try switching to PyPI...

h-vetinari:
@conda-forge/boost-cpp 
scipy now depends not on a prepackaged "boostinator", but on https://github.com/scipy/boost-headers-only/tree/2110ce20e59917b85b12059b3ddd8b133549f662 through [`scipy/_lib`](https://github.com/scipy/scipy/tree/v1.7.0rc1/scipy/_lib) - would it make sense to build this against conda-forge boost?

Besides the philosophical aspect of the question, there also seem to be build errors stemming from boost, e.g.
```
  scipy/stats/_boost/src/beta_ufunc.cxx:639:10: fatal error: /home/tyler/github_projects/scipy/scipy/stats/_boost/include/func_defs.hpp: No such file or directory
    639 | #include "/home/tyler/github_projects/scipy/scipy/stats/_boost/include/func_defs.hpp"
        |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  compilation terminated.
```

Though I have to say, that path looks a bit suspect as well @tylerjereddy

rgommers:
> would it make sense to build this against conda-forge boost?

What would the advantage be? SciPy requires this exact version of Boost (1.76.0 didn't work), it may or may not be patched in the future, and SciPy only supports building it via that submodule. 

> Though I have to say, that path looks a bit suspect as well @tylerjereddy

Yep, `scipy/stats/_boost/src/` is generated code and should be excluded from the sdist. That's a serious bug.

rgommers:
Should be fixed by https://github.com/scipy/scipy/pull/14196

h-vetinari:
Still failing with that patch - but then, it might be that this requires a newer sdist? I guess I could try deleting the stuff in there, OTOH, the failure on linux doesn't look related:
```
  scipy/stats/biasedurn.cxx:641:10: fatal error: numpy/random/bitgen.h: No such file or directory
    641 | #include "numpy/random/bitgen.h"
        |          ^~~~~~~~~~~~~~~~~~~~~~~
  compilation terminated.
```

on windows, the failure looks pythran-related.

h-vetinari:
On windows there's some lovely template spew:
```
  C:\bld\scipy_1623013654772\_h_env\lib\site-packages/pythran\pythonic/include/types/ndarray.hpp(495): error C2559: 'std::enable_if<`anonymous-namespace'::pythonic::types::is_numexpr_arg<F>::value&&std::is_same<bool,F::dtype>::value&&F::value!=&&!`anonymous-namespace'::pythonic::types::is_pod_array<F>::value,`anonymous-namespace'::pythonic::types::numpy_vexpr<`anonymous-namespace'::pythonic::types::ndarray<double,`anonymous-namespace'::pythonic::types::pshape<long>>,`anonymous-namespace'::pythonic::types::ndarray<long,`anonymous-namespace'::pythonic::types::pshape<long>>>>::type `anonymous-namespace'::pythonic::types::ndarray<double,`anonymous-namespace'::pythonic::types::pshape<long>>::operator [](const F &) const': cannot overload a member function without ref-qualifier with a member function with ref-qualifier
```

CC @serge-sans-paille

rgommers:
> Still failing with that patch - but then, it might be that this requires a newer sdist?

Indeed

> `fatal error: numpy/random/bitgen.h: No such file or directory`

This is numpy version related. It shouldn't try to use the `numpy.random` Cython API for numpy < 1.19.0 (see `isNPY_OLD` in `scipy/stats/_generate_pyx.py`). We're compiling against 1.17 here, so looks like something is wrong in the version detection or codegen.

rgommers:
> We're compiling against 1.17 here, so looks like something is wrong in the version detection or codegen.

Ah same problem, we're including the generated file in the sdist.

mckib2:
> What would the advantage be? SciPy requires this exact version of Boost (1.76.0 didn't work), it may or may not be patched in the future, and SciPy only supports building it via that submodule.

The only blocker to using a more recent Boost version is a GCC bug fixed in version 7.  ASAIK, it's only the backwards compatibility to GCC 4.8/5 that prevents updating

```
  scipy/stats/biasedurn.cxx:641:10: fatal error: numpy/random/bitgen.h: No such file or directory
    641 | #include "numpy/random/bitgen.h"
        |          ^~~~~~~~~~~~~~~~~~~~~~~
  compilation terminated.
```

I think the above is not pythran related, but rather complaining about unavailability of the newer numpy random API.  Is the numpy version consistent?

h-vetinari:
Update (aside from the other issues):
Arch builds pending on https://github.com/conda-forge/xsimd-feedstock/pull/72 & https://github.com/conda-forge/pythran-feedstock/pull/46

osx-arm builds pending on https://github.com/conda-forge/pythran-feedstock/pull/47

rgommers:
> > 
> 
> Ah same problem, we're including the generated file in the sdist.

Actually the needed solution is different. We still ship generated C/C++ files from Cython code. The problem is the use of an absolute path in an include statement. That should be replaced by using the correct include dir in the `add_extension` call. Would you be able to make that change @mckib2?

rgommers:
This `biasedurn` issue is a little more complicated than I first thought. Opened an issue: https://github.com/scipy/scipy/issues/14199

h-vetinari:
So, at least `pythran` is no available for all builds. As soon as theres a new rc, I'm happy to test this again.

rgommers:
Working on resolving the issues today.

h-vetinari:
OK, I had another idea to include the submodule (trivial in retrospect 🤦), which avoids the sdist issues.

While linux & osx are passing for cpython now, windows still fails to build with some pythran-related errors (`error C2559`) - could you have a look please, @serge-sans-paille?

Also, on PyPy, the test suite managed to crash all 8 pytest-xdist workers (both for linux & osx) with:
```
Stack (most recent call first, approximate line numbers):
  File "/home/[...]/site-packages/scipy/stats/_qmc.py", line 1019 in __init__
  File "/home/[...]/site-packages/scipy/stats/tests/test_stats.py", line 6886 in TestNumericalInverseHermite
  File "/home/[...]/site-packages/scipy/stats/tests/test_stats.py", line 1 in <module>
  [...]
  File "/home/[...]/site-packages/execnet/gateway_base.py", line 1552 in serve
  File "<string>", line 1 in <module>
  File "<string>", line 1 in <module>
Fatal Python error:   File "<builtin>/app_main.py", line 108 in run_toplevel
  File "<builtin>/app_main.py", line 606 in run_command_line
  File "<builtin>/app_main.py", line 940 in entry_point
Aborted
```

CC @mattip

recipe/meta.yaml
```diff
@@ -1,12 +1,17 @@
...
+{% set version = "1.7.0rc1" %}
 
 package:
   name: scipy
   version: {{ version }}
 
 source:
-  url: https://github.com/scipy/scipy/archive/v{{ version }}.tar.gz
-  sha256: 6cdb9d3aabe46c1602d9fd2f34812700068eddfe3ffd1dafcbd7e7960a874e1c
+  - url: https://github.com/scipy/scipy/archive/v{{ version }}.tar.gz
```
rgommers, line 8: Picking the GitHub-generated tarball seems unusual to me. The actual sdist is also there: https://github.com/scipy/scipy/releases/download/v1.7.0rc1/scipy-1.7.0rc1.tar.gz

rgommers:
> windows still fails to build with some pythran-related errors

It needs building with `clang-cl`, see https://github.com/scipy/scipy/blob/master/azure-pipelines.yml#L300.

And for the 32-bit Windows build, add `CXXFLAGS="-m32"` and `LDFLAGS="-m32"`.

mattip:
The PyPy abort seems to happen in the relatively new qmc code. I still don't see exactly what. Could the tests be run without xdist to see if that make the failure more clear?

rgommers:
> The PyPy abort seems to happen in the relatively new qmc code. 

Cc @tupui for that.

recipe/meta.yaml
h-vetinari: Well, that's just scipy being a exemplary FOSS citizen 😅 - most libraries don't have more than the auto-generated GH tarballs (which are still considered an authoritative source). In any case, assuming that sdist is the same as what was uploaded to PyPI, that would fail with the same problems as above. But I don't mind switching to this link for rc2.

h-vetinari:
> It needs building with `clang-cl`, see https://github.com/scipy/scipy/blob/master/azure-pipelines.yml#L300.

Thanks - the installation here is still [mainly](https://github.com/conda-forge/scipy-feedstock/blob/master/recipe/bld.bat#L30) `%PYTHON% -m pip install . -vv` - I presume `clang-cl` get's picked up if it's present?

> And for the 32-bit Windows build, add `CXXFLAGS="-m32"` and `LDFLAGS="-m32"`.

Conda-forge does not build for 32bit windows since several years...

h-vetinari:
> 
> 
> The PyPy abort seems to happen in the relatively new qmc code. I still don't see exactly what. Could the tests be run without xdist to see if that make the failure more clear?

Thanks for stopping by - pushed a commit to run with only one worker (though xdist is still installed; can remove if that turns out to be necessary as well)

rgommers:
>  I presume `clang-cl` get's picked up if it's present?

Yes it should, assuming it's on the PATH. On SciPy CI we install it with Chocolatey, and then add `set PATH=C:\Program Files\LLVM\bin;%PATH%`

serge-sans-paille:
@h-vetinari  The conda `clang` package installs its binaries as `Library/bin/clang-cl.exe` which is probably not in the default path (?)

h-vetinari:
@mattip @tupui
the crashes in the qmc code on pypy persist also without xdist.

recipe/meta.yaml
```diff
@@ -24,6 +29,8 @@ requirements:
     # numpy.distutils has a complex mechanism to facilitate mixing gfortran and MSVC
     # https://pav.iki.fi/blog/2017-10-08/pywingfortran.html#building-python-wheels-with-fortran-for-windows
     - {{ compiler('m2w64_fortran') }}   # [win]
+    # pythran-build needs clang-cl on windows
+    - clang                             # [win]
```
xhochy, line 33: You could try using the activation for `clang` on Windows to have all the environment variables correctly setup. 
```suggestion
    - {{ compiler('clang') }}                             # [win]
```

h-vetinari: Thanks, didn't know that the activation isn't done fully by `clang` itself already...

h-vetinari:
Hm... Thanks to @xhochy's tip, `clang-cl` is now being recognized, but it runs into something else I was worrying about - the setup here between gfortran & msvc is very finicky, and adding clang on top seems to break things:

```
  C:\bld\scipy_1623661568905\_build_env\Library\mingw-w64\bin\gfortran.exe -Wall -g -Wall -g -shared -nostdlib -Xclang --dependent-lib=msvcrt -fuse-ld=lld -Wl,-defaultlib:C:bldscipy_1623661568905_build_env/lib/clang/11.0.0/lib/windows/clang_rt.builtins-x86_64.lib build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqag.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqage.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagi.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagie.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagp.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagpe.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqags.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagse.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawc.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawce.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawf.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawfe.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawo.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawoe.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqaws.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawse.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqc25c.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqc25f.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqc25s.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqcheb.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqelg.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk15.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk15i.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk15w.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk21.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk31.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk41.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk51.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk61.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqmomo.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqng.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqpsrt.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqwgtc.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqwgtf.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqwgts.o build\temp.win-amd64-3.7\scipy\integrate\mach\d1mach.o build\temp.win-amd64-3.7\scipy\integrate\mach\xerror.o ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\blas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\cblas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\lapack.lib ..\..\..\..\..\..\bld\scipy_1623661568905\work\wrap_g77_abi_c.obj ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\blas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\cblas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\lapack.lib ..\..\..\..\..\..\bld\scipy_1623661568905\work\wrap_g77_abi_c.obj ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\blas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\cblas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\lapack.lib ..\..\..\..\..\..\bld\scipy_1623661568905\work\wrap_g77_abi_c.obj -LC:\bld\scipy_1623661568905\_build_env\Library\mingw-w64\lib\gcc\x86_64-w64-mingw32\5.3.0 -LC:\bld\scipy_1623661568905\_h_env\libs -LC:\bld\scipy_1623661568905\_h_env\PCbuild\amd64 -Lbuild\temp.win-amd64-3.7 -o build\temp.win-amd64-3.7\Release\.libs\libdqag.JBOONOHAOX4FH5ZITIT4TRISBKRGB7HY.gfortran-win_amd64.dll -Wl,--allow-multiple-definition -Wl,--output-def,build\temp.win-amd64-3.7\Release\libdqag.JBOONOHAOX4FH5ZITIT4TRISBKRGB7HY.gfortran-win_amd64.def -Wl,--export-all-symbols -Wl,--enable-auto-import -static -mlong-double-64
  gfortran.exe: error: unrecognized command line option '-Xclang'
  gfortran.exe: error: unrecognized command line option '--dependent-lib=msvcrt'
  gfortran.exe: error: unrecognized command line option '-fuse-ld=lld'
  error: Command "C:\bld\scipy_1623661568905\_build_env\Library\mingw-w64\bin\gfortran.exe -Wall -g -Wall -g -shared -nostdlib -Xclang --dependent-lib=msvcrt -fuse-ld=lld -Wl,-defaultlib:C:bldscipy_1623661568905_build_env/lib/clang/11.0.0/lib/windows/clang_rt.builtins-x86_64.lib build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqag.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqage.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagi.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagie.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagp.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagpe.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqags.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagse.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawc.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawce.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawf.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawfe.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawo.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawoe.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqaws.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawse.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqc25c.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqc25f.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqc25s.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqcheb.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqelg.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk15.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk15i.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk15w.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk21.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk31.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk41.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk51.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk61.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqmomo.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqng.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqpsrt.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqwgtc.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqwgtf.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqwgts.o build\temp.win-amd64-3.7\scipy\integrate\mach\d1mach.o build\temp.win-amd64-3.7\scipy\integrate\mach\xerror.o ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\blas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\cblas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\lapack.lib ..\..\..\..\..\..\bld\scipy_1623661568905\work\wrap_g77_abi_c.obj ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\blas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\cblas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\lapack.lib ..\..\..\..\..\..\bld\scipy_1623661568905\work\wrap_g77_abi_c.obj ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\blas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\cblas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\lapack.lib ..\..\..\..\..\..\bld\scipy_1623661568905\work\wrap_g77_abi_c.obj -LC:\bld\scipy_1623661568905\_build_env\Library\mingw-w64\lib\gcc\x86_64-w64-mingw32\5.3.0 -LC:\bld\scipy_1623661568905\_h_env\libs -LC:\bld\scipy_1623661568905\_h_env\PCbuild\amd64 -Lbuild\temp.win-amd64-3.7 -o build\temp.win-amd64-3.7\Release\.libs\libdqag.JBOONOHAOX4FH5ZITIT4TRISBKRGB7HY.gfortran-win_amd64.dll -Wl,--allow-multiple-definition -Wl,--output-def,build\temp.win-amd64-3.7\Release\libdqag.JBOONOHAOX4FH5ZITIT4TRISBKRGB7HY.gfortran-win_amd64.def -Wl,--export-all-symbols -Wl,--enable-auto-import -static -mlong-double-64" failed with exit status 1
  ERROR: Failed building wheel for scipy
```

mattip:
It seems PyPy has something wrong, I have it down to
```
>>> from scipy.stats import qmc
>>> sampler = qmc.Sobol(d=2, scramble=False)
pypy: unicodeobject.c:226: PyPyUnicode_FromFormatV: Assertion `((PyASCIIObject *)(obj))->wstr' failed.
```

h-vetinari:
A note regarding PPC: the travis builds were already flirting with timeouts (50min) before, and that despite being run with a reduced test suite (only scipy.linalg & scipy.fft tests). Now it seemed that the travis builds already hit that timeout during build time, so not much that can be done except changing CI...

I switched the PPC builds to azure (where they take much longer due to QEMU, but at least they run), and this runs into the same problems as when I tried it in #168 -  a gigantic number of failures:
```
= 2104 failed, 30393 passed, 2087 skipped, 107 xfailed, 12 xpassed, 961 warnings, 5 errors in 6275.90s (1:44:35) =
```

There seem to be a lot of float32 vs float64 problems, e.g.:
```
        A          = <3x4 sparse matrix of type '<class 'numpy.float64'>'
	with 5 stored elements in Compressed Sparse Column format>
        C          = <3x4 sparse matrix of type '<class 'numpy.float32'>'
	with 5 stored elements in Compressed Sparse Column format>
        self       = <scipy.sparse.tests.test_base.TestCSC object at 0x40809a1430>
        self_format = 'csc'
```

but some failures look really bad numerically (e.g. all zero where non-zero values are expected):
```
E   AssertionError: 
E   Arrays are not equal
E   
E   Mismatched elements: 9 / 24 (37.5%)
E   Max absolute difference: 134.
E   Max relative difference: 1.
E    x: matrix([[0, 0, 0, 0, 0, 0],
E           [0, 0, 0, 0, 0, 0],
E           [0, 0, 0, 0, 0, 0],
E           [0, 0, 0, 0, 0, 0]], dtype=int32)
E    y: matrix([[ -24.,  104.,    0.,   36.,    0.,    0.],
E           [  50., -126.,   10.,    0.,    0.,    0.],
E           [   0.,    0.,   -2.,   16.,    0.,    0.],
E           [   0.,    0.,  134.,    0.,    0.,    0.]])
        self       = <scipy.sparse.tests.test_base.TestCSR object at 0x4080b3c250>
```

Roughly half the errors seem to be coming from `linalg/tests/test_lapack.py`, so might be related to openblas. Though note that the linalg tests used to run through on travis (for 1.6.x), and I really don't understand what the relevant difference is here between azure & travis.

xhochy:
> Hm... Thanks to @xhochy's tip, `clang-cl` is now being recognized, but it runs into something else I was worrying about - the setup here between gfortran & msvc is very finicky, and adding clang on top seems to break things:
> 
> ```
>   C:\bld\scipy_1623661568905\_build_env\Library\mingw-w64\bin\gfortran.exe -Wall -g -Wall -g -shared -nostdlib -Xclang --dependent-lib=msvcrt -fuse-ld=lld -Wl,-defaultlib:C:bldscipy_1623661568905_build_env/lib/clang/11.0.0/lib/windows/clang_rt.builtins-x86_64.lib build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqag.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqage.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagi.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagie.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagp.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagpe.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqags.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagse.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawc.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawce.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawf.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawfe.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawo.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawoe.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqaws.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawse.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqc25c.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqc25f.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqc25s.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqcheb.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqelg.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk15.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk15i.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk15w.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk21.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk31.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk41.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk51.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk61.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqmomo.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqng.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqpsrt.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqwgtc.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqwgtf.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqwgts.o build\temp.win-amd64-3.7\scipy\integrate\mach\d1mach.o build\temp.win-amd64-3.7\scipy\integrate\mach\xerror.o ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\blas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\cblas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\lapack.lib ..\..\..\..\..\..\bld\scipy_1623661568905\work\wrap_g77_abi_c.obj ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\blas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\cblas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\lapack.lib ..\..\..\..\..\..\bld\scipy_1623661568905\work\wrap_g77_abi_c.obj ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\blas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\cblas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\lapack.lib ..\..\..\..\..\..\bld\scipy_1623661568905\work\wrap_g77_abi_c.obj -LC:\bld\scipy_1623661568905\_build_env\Library\mingw-w64\lib\gcc\x86_64-w64-mingw32\5.3.0 -LC:\bld\scipy_1623661568905\_h_env\libs -LC:\bld\scipy_1623661568905\_h_env\PCbuild\amd64 -Lbuild\temp.win-amd64-3.7 -o build\temp.win-amd64-3.7\Release\.libs\libdqag.JBOONOHAOX4FH5ZITIT4TRISBKRGB7HY.gfortran-win_amd64.dll -Wl,--allow-multiple-definition -Wl,--output-def,build\temp.win-amd64-3.7\Release\libdqag.JBOONOHAOX4FH5ZITIT4TRISBKRGB7HY.gfortran-win_amd64.def -Wl,--export-all-symbols -Wl,--enable-auto-import -static -mlong-double-64
>   gfortran.exe: error: unrecognized command line option '-Xclang'
>   gfortran.exe: error: unrecognized command line option '--dependent-lib=msvcrt'
>   gfortran.exe: error: unrecognized command line option '-fuse-ld=lld'
>   error: Command "C:\bld\scipy_1623661568905\_build_env\Library\mingw-w64\bin\gfortran.exe -Wall -g -Wall -g -shared -nostdlib -Xclang --dependent-lib=msvcrt -fuse-ld=lld -Wl,-defaultlib:C:bldscipy_1623661568905_build_env/lib/clang/11.0.0/lib/windows/clang_rt.builtins-x86_64.lib build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqag.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqage.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagi.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagie.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagp.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagpe.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqags.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqagse.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawc.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawce.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawf.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawfe.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawo.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawoe.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqaws.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqawse.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqc25c.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqc25f.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqc25s.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqcheb.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqelg.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk15.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk15i.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk15w.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk21.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk31.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk41.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk51.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqk61.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqmomo.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqng.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqpsrt.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqwgtc.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqwgtf.o build\temp.win-amd64-3.7\scipy\integrate\quadpack\dqwgts.o build\temp.win-amd64-3.7\scipy\integrate\mach\d1mach.o build\temp.win-amd64-3.7\scipy\integrate\mach\xerror.o ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\blas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\cblas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\lapack.lib ..\..\..\..\..\..\bld\scipy_1623661568905\work\wrap_g77_abi_c.obj ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\blas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\cblas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\lapack.lib ..\..\..\..\..\..\bld\scipy_1623661568905\work\wrap_g77_abi_c.obj ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\blas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\cblas.lib ..\..\..\..\..\..\bld\scipy_1623661568905\_h_env\Library\lib\lapack.lib ..\..\..\..\..\..\bld\scipy_1623661568905\work\wrap_g77_abi_c.obj -LC:\bld\scipy_1623661568905\_build_env\Library\mingw-w64\lib\gcc\x86_64-w64-mingw32\5.3.0 -LC:\bld\scipy_1623661568905\_h_env\libs -LC:\bld\scipy_1623661568905\_h_env\PCbuild\amd64 -Lbuild\temp.win-amd64-3.7 -o build\temp.win-amd64-3.7\Release\.libs\libdqag.JBOONOHAOX4FH5ZITIT4TRISBKRGB7HY.gfortran-win_amd64.dll -Wl,--allow-multiple-definition -Wl,--output-def,build\temp.win-amd64-3.7\Release\libdqag.JBOONOHAOX4FH5ZITIT4TRISBKRGB7HY.gfortran-win_amd64.def -Wl,--export-all-symbols -Wl,--enable-auto-import -static -mlong-double-64" failed with exit status 1
>   ERROR: Failed building wheel for scipy
> ```

That is because the `LDFLAGS` are setup for using `clang` and make no sense for `gfortran`. We could either:

- Try to link with `clang` here, too (not so optimistic this will work)
- Separate somehow fortran/non-fortran LDFLAGS.

Can someone point me to the Windows setup scipy is using in their CI? (Happy to help with really specific stuff but don't have the time to dig through the scipy code base, sadly)

h-vetinari:
> Can someone point me to the Windows setup scipy is using in their CI?

I think this diverged a bit already - windows support for scipy in conda-forge was a long-standing [issue](https://github.com/conda-forge/scipy-feedstock/issues/80) that got finally [solved](https://github.com/conda-forge/scipy-feedstock/pull/146) by @isuruf late last year, cf. also [this PR](https://github.com/conda-forge/scipy-feedstock/pull/140) (though it's my understanding that the longterm goal would be to use the new flang in LLVM, once it can start generating bytecode - probably LLVM 14/15).

rgommers:
Thanks @xhochy! For SciPy CI:

- Set up of clang-cl: https://github.com/scipy/scipy/blob/master/azure-pipelines.yml#L294
- Set up of env vars (only for 32-bit, for 64-bit `LDFLAGS` is not defined): https://github.com/scipy/scipy/blob/master/azure-pipelines.yml#L340

The wheel build setup in https://github.com/MacPython/scipy-wheels/blob/master/appveyor.yml is essentially the same.

rgommers:
> I think this diverged a bit already - windows support for scipy in conda-forge was a long-standing [issue](https://github.com/conda-forge/scipy-feedstock/issues/80) that got finally [solved](https://github.com/conda-forge/scipy-feedstock/pull/146) by @isuruf late last year, cf. also [this PR](https://github.com/conda-forge/scipy-feedstock/pull/140) (though it's my understanding that the longterm goal would be to use the new flang in LLVM, once it can start generating bytecode - probably LLVM 14/15).

What @isuruf did  there essentially replicates the glueing together of MSVC and Gfortran that's also done in `numpy.distutils`.

xhochy:
@h-vetinari Can you please try unsetting `LDFLAGS` on Windows? Maybe most things works without the `LDFLAGS`. Looking at the log, I don't see `clang-cl` being used though.

h-vetinari:
> Looking at the log, I don't see `clang-cl` being used though.

It is being used (in some way at least). Before it got added (and picked up by `pip install ...`), things were failing with a lot of template spew.

> Can you please try unsetting `LDFLAGS` on Windows?

Will do

h-vetinari:
> > Can you please try unsetting `LDFLAGS` on Windows?
> 
> Will do

Seems that this gets us through the fortran builds, but still fails once clang-cl should be used (for compiling pythran code)
```
  building 'scipy.interpolate._rbfinterp_pythran' extension
  compiling C sources
  clang++.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MD -DENABLE_PYTHON_MODULE -D__PYTHRAN__=3 -DPYTHRAN_BLAS_NONE -IC:\bld\scipy_1623680406520\_h_env\lib\site-packages/pythran -IC:\bld\scipy_1623680406520\_h_env\lib\site-packages\numpy\core\include -IC:\bld\scipy_1623680406520\_h_env\lib\site-packages\numpy\core\include -IC:\bld\scipy_1623680406520\_h_env\include -IC:\bld\scipy_1623680406520\_h_env\include -IC:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\ATLMFC\include -IC:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\include -IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.6.1\include\um -IC:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\ucrt -IC:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\shared -IC:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\um -IC:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\winrt -IC:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\cppwinrt -IC:\bld\scipy_1623680406520\_h_env\Library\include -IC:\bld\scipy_1623680406520\_h_env\Library\include /EHsc /Tpscipy/interpolate/_rbfinterp_pythran.cpp /Fobuild\temp.win-amd64-3.7\Release\scipy/interpolate/_rbfinterp_pythran.obj /std:c++14 -fvisibility-inlines-hidden -std=c++14 -fmessage-length=0 -march=k8 -mtune=haswell -ftree-vectorize -fstack-protector-strong -O2 -ffunction-sections -pipe -D_CRT_SECURE_NO_WARNINGS -D_MT -D_DLL -nostdlib -Xclang --dependent-lib=msvcrt -fuse-ld=lld -fno-aligned-allocation /Zm1000
  clang++: error: no such file or directory: '/c'
  clang++: error: no such file or directory: '/nologo'
  clang++: error: no such file or directory: '/Ox'
  clang++: error: no such file or directory: '/W3'
  clang++: error: no such file or directory: '/GL'
  clang++: error: no such file or directory: '/DNDEBUG'
  clang++: error: no such file or directory: '/MD'
  clang++: error: no such file or directory: '/EHsc'
  clang++: error: no such file or directory: '/Tpscipy/interpolate/_rbfinterp_pythran.cpp'
  clang++: error: no such file or directory: '/Fobuild\temp.win-amd64-3.7\Release\scipy/interpolate/_rbfinterp_pythran.obj'
  clang++: error: no such file or directory: '/std:c++14'
  clang++: error: no such file or directory: '/Zm1000'
  clang++: error: no input files
  error: Command "clang++.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MD -DENABLE_PYTHON_MODULE -D__PYTHRAN__=3 -DPYTHRAN_BLAS_NONE -IC:\bld\scipy_1623680406520\_h_env\lib\site-packages/pythran -IC:\bld\scipy_1623680406520\_h_env\lib\site-packages\numpy\core\include -IC:\bld\scipy_1623680406520\_h_env\lib\site-packages\numpy\core\include -IC:\bld\scipy_1623680406520\_h_env\include -IC:\bld\scipy_1623680406520\_h_env\include -IC:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\ATLMFC\include -IC:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\include -IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.6.1\include\um -IC:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\ucrt -IC:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\shared -IC:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\um -IC:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\winrt -IC:\Program Files (x86)\Windows Kits\10\include\10.0.17763.0\cppwinrt -IC:\bld\scipy_1623680406520\_h_env\Library\include -IC:\bld\scipy_1623680406520\_h_env\Library\include /EHsc /Tpscipy/interpolate/_rbfinterp_pythran.cpp /Fobuild\temp.win-amd64-3.7\Release\scipy/interpolate/_rbfinterp_pythran.obj /std:c++14 -fvisibility-inlines-hidden -std=c++14 -fmessage-length=0 -march=k8 -mtune=haswell -ftree-vectorize -fstack-protector-strong -O2 -ffunction-sections -pipe -D_CRT_SECURE_NO_WARNINGS -D_MT -D_DLL -nostdlib -Xclang --dependent-lib=msvcrt -fuse-ld=lld -fno-aligned-allocation /Zm1000" failed with exit status 1
  ERROR: Failed building wheel for scipy
```

recipe/meta.yaml
```diff
@@ -19,13 +24,15 @@ requirements:
     - cython                                 # [build_platform != target_platform]
     - numpy                                  # [build_platform != target_platform]
     - pybind11                               # [build_platform != target_platform]
-    - {{ compiler('fortran') }}   # [unix]
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    # pythran code needs clang on windows
+    - {{ compiler('clang') }}                # [win]
```
isuruf, line 30: Don't use this. This is not using `clang-cl`, but `clang`. Go back to just `- clang` and set env variables `CC` and `CXX` to `clang-cl`

xhochy: @isuruf Should we make a `compiler('clang-cl')` activation once things work here? I found it quite useful to have these reusable activation scripts.

h-vetinari: A couple commits ago, I just had `- clang`, but then `clang-cl` was not on the `PATH`. I assume I'll also need add something to the `PATH`?

isuruf: Nope. It should be on the PATH.

xhochy: https://github.com/conda-forge/clang-win-activation-feedstock/blob/724a3900c88ef1748a9c1d90d0c2927db99c3ef5/recipe/activate-clang_win-64.bat#L1 isn't adding anything to the `PATH`, so it must be that you are missing a dependency that is specified here: https://github.com/conda-forge/clang-win-activation-feedstock/blob/724a3900c88ef1748a9c1d90d0c2927db99c3ef5/recipe/meta.yaml#L41

isuruf: As I said earlier, none of those dependencies are needed. Just `clang`

serge-sans-paille:
This is using clang++ and no clang++-cl :-/

recipe/meta.yaml
h-vetinari: Just `- clang` does not add `clang-cl` to the PATH (the presence-check I added in `bld.bat` is failing):
```
(%BUILD_PREFIX%) %SRC_DIR%>clang-cl.exe --version 

(%BUILD_PREFIX%) %SRC_DIR%>if -1073741515 NEQ 0 exit 1 
```

h-vetinari:
Thanks @isuruf; I opened a PR for clangdev to fix the missing zlib.

Though the compilation of pythran-code here is still failing:
```
  building 'scipy.interpolate._rbfinterp_pythran' extension
  compiling C sources
  clang-cl /c /nologo /Ox [...opts, includes...] /d1trimfile:C:\bld\scipy_1623686394883\work [...other...]
  clang-cl: error: no such file or directory: '/d1trimfile:C:\bld\scipy_1623686394883\work'
  error: Command "clang-cl /c /nologo /Ox [...]" failed with exit status 1
  ERROR: Failed building wheel for scipy
```

`'/d1trimfile:C:\bld\scipy_1623686394883\work'` looks super weird - as if there's a space missing or so?

h-vetinari:
@rgommers @tylerjereddy
Windows builds [ran through](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=337091&view=results) for rc1 (thanks to the help of @isuruf & @xhochy), but now [fail](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=337360&view=results) on rc2:
```
=========================== short test summary info ===========================
FAILED integrate/_ivp/tests/test_ivp.py::test_integration_sparse_difference
FAILED optimize/tests/test__numdiff.py::test_group_columns - TypeError: Inval...
FAILED optimize/tests/test__numdiff.py::TestApproxDerivativeSparse::test_check_derivative
FAILED optimize/tests/test_least_squares.py::TestDogbox::test_numerical_jac
FAILED optimize/tests/test_least_squares.py::TestDogbox::test_with_bounds - T...
FAILED optimize/tests/test_least_squares.py::TestTRF::test_numerical_jac - Ty...
FAILED optimize/tests/test_least_squares.py::TestTRF::test_with_bounds - Type...
= 7 failed, 32021 passed, 2565 skipped, 105 xfailed, 10 xpassed, 41 warnings in 468.41s (0:07:48) =
```

As far as I can see, all share the same underlying error:
```
E   TypeError: Invalid call to pythranized function `group_sparse(int, int, intc[:] (reshaped), intc[:])'
E   Candidates are:
E   
E       - group_sparse(int, int, int[:], int[:])
E       - group_sparse(int, int, int32[:], int32[:])
```

CC @serge-sans-paille

rgommers:
There is no relevant difference between `rc1` and `rc2` in the Python/Pythran code. The difference is that you're now building from the sdist, which has pre-compiled Pythran sources. I checked the contents of the file, and they're generated with Pythran `0.9.11` which is correct. So not sure why that would be different from using the conda-forge Pythran package at build time.

The other difference is that for `rc1` we included Cython sources and for `rc2` we didn't anymore. `_group_columns` has a Cython source as well, but that should only be used if `PYTHRAN_USE_SCIPY` is explicitly set to `0`.

h-vetinari:
> The difference is that you're now building from the sdist, which has pre-compiled Pythran sources.

Ah, I was under the impression that the sdist was going to remove all that precompiled code. That's IMO an argument to stay with the more explicit raw-sources-plus-submodule setup. Let me try if that removes the failures.

rgommers:
> Ah, I was under the impression that the sdist was going to remove all that precompiled code. That's IMO an argument to stay with the more explicit raw-sources-plus-submodule setup. Let me try if that removes the failures.

For Cython it did, because we're comfortable now requiring Cython as a build-time dependency. We don't want to require everyone to have a specific version of Pythran installed though, we need to get some more experience with it first.

h-vetinari:
> For Cython it did, because we're comfortable now requiring Cython as a build-time dependency. We don't want to require everyone to have a specific version of Pythran installed though, we need to get some more experience with it first.

OK, makes sense. In this case, I think it makes more sense for conda-forge to build from scratch, because pythran is easily available here, especially as windows seems to fail with the precompiled sources.

tylerjereddy:
Are there enough concerns that we may need an `rc3` ? At the moment, I don't see any concrete suggestions for specific additional backports that are needed?

rgommers:
> Are there enough concerns that we may need an `rc3` ? At the moment, I don't see any concrete suggestions for specific additional backports that are needed?

It looks like we don't. Current status seems to be:
- Windows, Linux, macOS builds pass
- PyPy build is crashing, and there's lots of random warnings in the build log
- ppc64le build worked, but tests are timing out
- aarch64 build is timing out

We do have wheel builds for aarch64, so there's no fundamental problem there. Not sure what to do to not make the builds time out, we didn't change the SciPy build times _that_ much in this release (main addition is the use of Boost probably, I'd guess ~10% increase total).

For the PyPy builds, it's crashing in the test collection phase after a ton of warnings in the build phase:
```
Warning: rpath /home/conda/feedstock_root/build_artifacts/scipy_1623780293469/_build_env/lib is outside prefix /home/conda/feedstock_root/build_artifacts/scipy_1623780293469/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place (removing it)
...
INFO (scipy,lib/python3.7/site-packages/scipy/stats/_boost/nbinom_ufunc.pypy37-pp73-x86_64-linux-gnu.so): Needed DSO x86_64-conda-linux-gnu/sysroot/lib64/libc.so.6 found in CDT/compiler package conda-forge::sysroot_linux-64-2.12-h77966d4_13
...
2021-06-15T18:27:20.8955769Z WARNING (scipy): plugin library (Python) package conda-forge::numpy-1.19.5-py37h620df1f_1 in requirements/run but it is not used (i.e. it is overdepending or perhaps statically linked? If that is what you want then add it to `build/ignore_run_exports`)
2021-06-15T18:27:20.8957053Z WARNING (scipy): dso library package conda-forge::pypy3.7-7.3.4-hc4e864a_4 in requirements/run but it is not used (i.e. it is overdepending or perhaps statically linked? If that is what you want then add it to `build/ignore_run_exports`)
2021-06-15T18:27:20.8958305Z WARNING (scipy): interpreter (Python) package conda-forge::python-3.7.10-0_73_pypy in requirements/run but it is not used (i.e. it is overdepending or perhaps statically linked? If that is what you want then add it to `build/ignore_run_exports`)
```
not sure if the last warnings here are relevant or not. Either way, I don't think PyPy should be blocking. We have no PyPy CI on the SciPy repo, so this could be anything.

h-vetinari:
Largely agree with @rgommers, no extra rc is necessary.

* PyPy on x86 was having issues that were being [looked at](https://github.com/conda-forge/scipy-feedstock/pull/169#issuecomment-860648391) by @mattip - don't know the current status, but we can carry a patch if necessary once a solution becomes available.
* PPC builds were flirting with timeouts before on Travis, and are now consistently timing out, often already during the build phase (though I did see a single [passing](https://travis-ci.com/github/conda-forge/scipy-feedstock/builds/229020274) job)
  * It would be possible to run those jobs on Azure through QEMU, where the build runs through, but then ~2000 tests fail (see [this comment](https://github.com/conda-forge/scipy-feedstock/pull/169#issuecomment-860688710) for more details)
  * The availability of wheels does not help the conda-forge side - unless conda-forge/core manually builds them according to CFEP-13 (or exceptionally allows binary repackaging).
* aarch was mostly fine, just individual builds timing out if they hit the slower aarch machine, which can be solved by restarts.

The one thing I could think of to perhaps slightly ameliorate the timeout situation is to use the sdist with the pre-compiled pythran sources for aarch & ppc. Note that we cannot do this universally because it [fails](https://github.com/conda-forge/scipy-feedstock/pull/169#issuecomment-861350431) on windows (seems like the generated pythran sources are not fully portable).

rgommers:
> Note that we cannot do this universally because it [fails](https://github.com/conda-forge/scipy-feedstock/pull/169#issuecomment-861350431) on windows (seems like the generated pythran sources are not fully portable).

Oh yes, we need to keep track of that, and may need to fix that before the release. Issue filed at https://github.com/serge-sans-paille/pythran/issues/1811.

h-vetinari:
> Oh yes, we need to keep track of that, and may need to fix that before the release.

I agree it's good to fix (and commented on the issue), just not strictly necessary for conda-forge. I'd prefer compiling from source completely, including pythran

h-vetinari:
This is at the same point where it was for rc2 - ppc is consistently timing out, pypy+x86 is failing, the rest is more or less OK (occasional timeouts in drone depending on the machine).

If people want, I'd be willing to join as a maintainer and babysit the CI by triggering restarts until the builds are uploaded.

rgommers:
> This is at the same point where it was for rc2 - ppc is consistently timing out, pypy+x86 is failing, the rest is more or less OK (occasional timeouts in drone depending on the machine).

SciPy builds got heavier, on the main repo we're getting regular timeouts on our Azure Windows build + full test suite as well. Working on improving build times, but I think there's also scope to add caching better between CI runs (e.g., use `ccache` and persist it between runs).

Looks like there's no point restarting `ppc64le` here? Drone wants too many permissions, so I'm hesitating to give it access with my personal account.

> If people want, I'd be willing to join as a maintainer and babysit the CI by triggering restarts until the builds are uploaded.

Thanks @h-vetinari, +1 from me! Want to send a separate PR to add yourself?

h-vetinari:
> SciPy builds got heavier, on the main repo we're getting regular timeouts on our Azure Windows build + full test suite as well. Working on improving build times, but I think there's also scope to add caching better between CI runs (e.g., use `ccache` and persist it between runs).

I don't think the caching would be so easy to pair with the current conda-forge infrastructure.

> Looks like there's no point restarting `ppc64le` here? Drone wants too many permissions, so I'm hesitating to give it access with my personal account.

As I mentioned [above](https://github.com/conda-forge/scipy-feedstock/pull/169#issuecomment-863844147), there was at least one passing PPC build. I've also re-checked the diff since then and am retrying if reverting https://github.com/conda-forge/scipy-feedstock/pull/169/commits/052cff0f3fd5997c7bb0fecc25265e9d1798ec6e helps.

I gave drone those rights, and have been able to restart builds on other feedstocks where I"m the maintainer.

> Thanks @h-vetinari, +1 from me! Want to send a separate PR to add yourself?

Cool, happy to hear it! I took the liberty to add this to it here - it can wait until this PR is merged.

h-vetinari:
> @mattip: It seems PyPy has something wrong, I have it down to
> 
> ```
> >>> from scipy.stats import qmc
> >>> sampler = qmc.Sobol(d=2, scramble=False)
> pypy: unicodeobject.c:226: PyPyUnicode_FromFormatV: Assertion `((PyASCIIObject *)(obj))->wstr' failed.
> ```

Hey @mattip, no rush, just checking if there's been any news on this since last week.

mattip:
I haven't been able to make progress with the PyPy error. Is there a way to release the other builds but hold PyPy back?

h-vetinari:
> 
> 
> I haven't been able to make progress with the PyPy error. Is there a way to release the other builds but hold PyPy back?

The builds will not be uploaded if the CI fails, so this would be mergeable as is without issue.

rgommers:
I had a look at the Drone timeout, it loses almost 10 minutes on a conda solve:

image

And it was not even close to done - it timed out right after starting on `sparsetools`, which is one extension that takes about ~20% of the total build time.

rgommers (approved):
Changes LGTM, and Windows, macOS and Linux builds are green. So let's get this in. 

Thanks @h-vetinari and everyone else!

h-vetinari:
Update: all aarch builds are [available](https://anaconda.org/conda-forge/scipy/files) now. For ppc, one build ran through (py39), been restarting the others but with no success so far.

Note that ppc+pypy always stalls sometime well before the 50min cutoff, à la:
```
The following NEW packages will be INSTALLED:
    _libgcc_mutex:      0.1-conda_forge            conda-forge
    _openmp_mutex:      4.5-1_gnu                  conda-forge
    attrs:              21.2.0-pyhd8ed1ab_0        conda-forge
    bzip2:              1.0.8-h4e0d66e_4           conda-forge
    ca-certificates:    2021.5.30-h1084571_0       conda-forge
    execnet:            1.9.0-pyhd8ed1ab_0         conda-forge
    expat:              2.4.1-h3b9df90_0           conda-forge
    gdbm:
    
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated
```

rgommers:
Nice, thanks @h-vetinari. For the aarch64 builds, is it just getting scheduled on faster hardware? Would be nice to see the build log or a summary on why it worked now.

h-vetinari:
> 
> 
> Nice, thanks @h-vetinari. For the aarch64 builds, is it just getting scheduled on faster hardware? Would be nice to see the build log or a summary on why it worked now.

AFAIU, there are two different machines that make up the drone queue, one larger and one smaller - the difference in build time can be quite substantial (~50%), which obviously becomes relevant if the builds are starting to time out on the smaller machine. You can see some passing builds after the restart [here](https://cloud.drone.io/conda-forge/scipy-feedstock/242).

h-vetinari:
PPC + cpython are now built as well, I've stopped retriggering PPC + PyPy because it reproducibly leads to hangs.

That means 1.7.0 has now been built for all but the following three variants:
* pypy37 on linux / x64
* pypy37 on linux / ppc
* pypy37 on osx / x64

rgommers:
Thanks for shepherding those builds @h-vetinari!

mattip:
If I can figure out what is going on with PyPy + Cython + SciPy, I will create a new PR.

mattip:
I think the PyPy crash is due to a huge stack allocation when entering a function, cython/cython#4272. I don't understand what the difference is between CPython and PyPy.

Issue #173, opened 2021-07-01
3 missing pypy builds for 1.7.0


h-vetinari:
See discussion in #169, culminating in [this comment](https://github.com/conda-forge/scipy-feedstock/pull/169#issuecomment-869601570): 

> That means 1.7.0 has now been built for all but the following three variants:
>   * pypy37 on linux / x64
>   * pypy37 on linux / ppc
>   * pypy37 on osx / x64

mattip:
I think the PyPy crash is due to a huge stack allocation when entering a function, cython/cython#4272. I don't understand what the difference is between CPython and PyPy.

h-vetinari:
The PPC build will need some encouragement (=restarts), keeping this open until then.

mattip:
I think windows64 pypy migration is missing as well?

h-vetinari:
Yes indeed, but I wasn't counting that since the migrator still needs to chew through a bunch of dependencies first

mattip:
Is the migrator still active? At https://conda-forge.org/status/ I see it is listed under "Recently closed migrations" as "pypy37-windows closed at 2021-07-07"

h-vetinari:
See my comment [here](https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/1668#issuecomment-882343614)

h-vetinari:
I cannot get the pypy build on PPC to pass despite >5 restarts. Perhaps it needs some of the recent fixes on the pypy-feedstock (and PPC needing manual rebuilds)

mattip:
I personally would be happy to drop ppc and s390x support for PyPy on conda forge. At least until we get a little more experience around the use case for PyPy on x86/arm64.

isuruf:
travis is supposed to run only a fraction as in https://github.com/conda-forge/scipy-feedstock/blob/master/recipe/run_test.py#L19-L21. Not sure why that doesn't get triggered.

h-vetinari:
@isuruf, it's not a question of runtime. The build regularly fails with:
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
```

[...]

PR #179, opened 2021-07-22
Try cross-compiling ppc&aarch

h-vetinari:
See discussion in #173. This [previously](https://github.com/conda-forge/scipy-feedstock/pull/169#issuecomment-860688710) ran into troubles, so I'm not expecting this to pass, unfortunately.


h-vetinari:
I had started this before my holidays, surprised to see that now PPC builds ran through! :)

h-vetinari:
@conda-forge/scipy @isuruf
This builds the first osx-arm builds for 1.7.1 (needed new pythran release which happened recently) - this is unrelated to this PR, but just needed a rerun.

More to the (original) point of this PR, despite #185, drone kept timing out, so I'm back to cross-compilation (or, more accurately: emulation of aarch through qemu on x86).

The runtime is pretty long, but - to my mind - still better than having to constantly restart failed jobs.

Personally I would merge this PR as-is, but wanted to get some feedback.

h-vetinari:
Happy to hear!

Regarding precedent, there's quite a number of packages that do this, so this is not breaking any new ground. Due to emulation, it's generally slower than drone, but in some cases can still be better.

PR #184, opened 2021-08-06
Separate tests into their own package

### LINKED ISSUE #160: Separate tests into a `scipy-tests` package
- [x] I read [the conda-forge documentation](https://conda-forge.org/docs/user/introduction.html#how-can-i-install-packages-from-conda-forge) and could not find the solution for my problem there.

The tests in the scipy package are quite large (11MiB on-disk size). Thus it would be nice if we could ship them in a separate conda package like we did for `pyarrow`: https://github.com/conda-forge/arrow-cpp-feedstock/blob/48de7951f41b685cd1112e2f85c7c016eac0f553/recipe/meta.yaml#L232 This would allow downstream consumers to still run the `scipy` tests in their environment but allow those that don't want to do this to not ship the tests at all.

h-vetinari:
Continuing @xhochy's work from #175, ~lightly edited.~

Fixes #160 
Closes #175


h-vetinari:
@xhochy, is there a way to avoid the full rebuild for the tests? Could we not make a top-level script that runs only once and copies then copy the right files from build folder as needed?

With the current approach, we're doubling CI time, which is a no-go on drone, and potentially critical on stuff like cross-compiling PPC.

h-vetinari:
@xhochy @rgommers 
Made some good progress finally on this old PR. It builds on unix now, where the new meson build gives us a nice separation into the global `build.sh` phase and then installing the correct files for the respective outputs.

Windows is harder to do with the distutils build (no good separation into build & installation steps, so have to do a full recompilation twice; this then fails for reasons that elude me currently, perhaps due to not cleaning up some build artefacts from the previous run correctly; any insights there would be appreciated).

In any case, it should drastically reduce the default download size, so should be a nice win and is probably worth doing even if it's unix-only for now.

recipe/build-output.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+set -ex
+
+# Set a few environment variables that are not set due to
+# https://github.com/conda/conda-build/issues/3993
+export PIP_NO_BUILD_ISOLATION=False
```
rgommers, line 6: Shouldn't this be `True`? We never want build isolation.

rgommers: Guess it doesn't matter because we already have a wheel, but it's potentially confusing.

recipe/meta.yaml
```diff
@@ -80,105 +73,190 @@ requirements:
...
+        {% set tests_to_skip = tests_to_skip + " or test_cdist_ and (refcount or calling_convention)" %}  # [pypy_emu]
+        {% set tests_to_skip = tests_to_skip + " or test_conditionally_positive_definite" %}              # [pypy_emu]
+        {% set tests_to_skip = tests_to_skip + " or test_cont_basic[500 or test_cython_api[elliprj]" %}   # [pypy_emu]
+        {% set tests_to_skip = tests_to_skip + " or test_filtfilt_gust or test_kappa3_array_gh13582" %}   # [pypy_emu]
+        {% set tests_to_skip = tests_to_skip + " or test_fitstart[genexpon-shapes28] or test_gh12922" %}  # [pypy_emu]
+        {% set tests_to_skip = tests_to_skip + " or test_integration_sparse or test_kdtree_box" %}        # [pypy_emu]
+        {% set tests_to_skip = tests_to_skip + " or test_minimize_callback_copies_array[fmin]" %}         # [pypy_emu]
+        {% set tests_to_skip = tests_to_skip + " or test_mip1 or test_mixed_threads_processes" %}         # [pypy_emu]
+        {% set tests_to_skip = tests_to_skip + " or test_modules_importable or test_neldermead_limit" %}  # [pypy_emu]
+        {% set tests_to_skip = tests_to_skip + " or test_random_exact or test_random_complex_exact" %}    # [pypy_emu]
```
rgommers, line 248: Some of these shouldn't be running I think because of the `set label = "'fast'"` logic a few lines below. For the ones that are actually needed, how about we upstream the skips?

rgommers (commented):
These test files aren't being deleted yet with the pattern being `*/tests/*`, that'd be another 800 kb or so.
```
_lib/_test_ccallback.cpython-39-darwin.so
_lib/_test_deprecation_call.cpython-39-darwin.so
_lib/_test_deprecation_def.cpython-39-darwin.so
integrate/_test_multivariate.cpython-39-darwin.so
integrate/_test_odeint_banded.cpython-39-darwin.so
io/_test_fortran.cpython-39-darwin.so
ndimage/_ctest.cpython-39-darwin.so
ndimage/_cytest.cpython-39-darwin.so
special/_test_internal.cpython-39-darwin.so
```

It'd be great to summarize the final gain here (total size with/without); it'll justify doing this and maybe in the future doing this for SciPy itself or rearranging the folder layout so it's much easier to do.

+1 for not doing this on Windows right now, doubling the build time isn't worth it and once we get rid of `distutils` it'll be easy to do.

This does still require a patch to ensure that `scipy.test()` or `pytest --pyargs scipy` clearly tells the user to install `scipy-tests`.

recipe/build-output.sh
h-vetinari: Yeah, good point, will fix.

recipe/meta.yaml
h-vetinari: All these are current (as in: running even with `label='fast'`). I updated them recently in https://github.com/conda-forge/scipy-feedstock/pull/227

h-vetinari:
> It'd be great to summarize the final gain here (total size with/without); it'll justify doing this and maybe in the future doing this for SciPy itself or rearranging the folder layout so it's much easier to do.

I don't see the current footprint but it'd show up once it's on https://anaconda.org/conda-forge/scipy/files. I'll update with a summary afterwards.

> These test files aren't being deleted yet with the pattern being */tests/*, that'd be another 800 kb or so. [...]

For now, I think I'll take the win of removing the test folders with a simple procedure. We can fine-tune afterwards IMO.

> +1 for not doing this on Windows right now, doubling the build time isn't worth it and once we get rid of `distutils` it'll be easy to do.

On windows I actually don't mind the doubled built time. We have long-running jobs for the emulated test suites on aarch/ppc anyway that run way longer than a twice compiled build on windows. But I don't know why it's failing to compile the second time (with identical build script & host deps), and if it becomes too vexing I might give up. 🤷 

> This does still require a patch to ensure that `scipy.test()` or `pytest --pyargs scipy` clearly tells the user to install `scipy-tests`.

Fair enough.

h-vetinari:
> > This does still require a patch to ensure that `scipy.test()` or `pytest --pyargs scipy` clearly tells the user to install `scipy-tests`.
> 
> Fair enough.

While I have an idea of how to patch `scipy.test`, I'm not actually sure how to do that for `pytest --pyargs foo`. What I can tell from the pytest [docs](https://docs.pytest.org/en/latest/how-to/usage.html) is that it will auto-discover the tests based on the filesystem location of `foo`.

Pytest has quite an array of [hooks](https://docs.pytest.org/en/latest/reference/reference.html#hooks), but I'm not even sure if any of scipy's code will run if invoked with `pytest --pyargs scipy`? Do you have an idea/intuition how to achieve that?

rgommers:
> I'm not even sure if any of scipy's code will run if invoked with `pytest --pyargs scipy`? Do you have an idea/intuition how to achieve that?

How about simply writing a file `scipy/_lib/test_condaforge.py` with a single test function that raises an exception saying to `conda install scipy-tests`?

h-vetinari:
> How about simply writing a file `scipy/_lib/test_condaforge.py` with a single test function that raises an exception saying to `conda install scipy-tests`?

I didn't think of that because I was aiming to not get to the test collection stage, but I doubt that's possible. So I think that's a pretty good idea because, not least because it solves both cases at once, and because we don't even have to patch it in, we can just copy it from `$RECIPE_DIR` to `$SP_DIR`.

h-vetinari:
So this comes with a few extra bells & whistles now 🙃 

> These test files aren't being deleted yet with the pattern being `*/tests/*`, that'd be another 800 kb or so.

> This does still require a patch to ensure that `scipy.test()` or `pytest --pyargs scipy` clearly tells the user to install `scipy-tests`.

Both of these are done and tested now. Also I've figured out windows (recompilation is not elegant, but better than not having the separation; I'll switch away from that as soon as we have meson builds on windows).

We can still upstream part of the pypy-in-emulation skips, but it's quite specific to that combination. For example, the slowest pypy tests on native hardware take ~120sec, the slowest cpython test in emulation take ~360sec, but for pypy-in-emulation, many tests will blow through our 1200sec timeout. Accurately representing that sounds like a hassle, but if we don't mind marking a couple tests as non-`'fast'`, then all the better of course.

In any case, this PR should be on home stretch now.

h-vetinari:
> Happy to take a final look when it's ready.

It is, that's what I meant by home stretch. :)

> marking things as `xslow` for PyPy only would be fully justified if they take 120 seconds without emulation.

I'll have a look at adding some markers. For context, I gave the numbers for the very longest running tests (according to the logs), not all of them take ~120sec without emulation, just 1-2.

h-vetinari:
> > Happy to take a final look when it's ready.
> 
> It is, that's what I meant by home stretch. :)

OK, I rebased this after #234 and managed to (hopefully finally) beat those intermittent errors on windows.

rgommers (approved):
Two timeouts for PyPy on `aarch64` it looks like.

This LGTM in principle. It'd be great if someone else could review too though, I can't really judge the `.bat` file changes or whether this is all according to conda-forge best practices. From the "how the `scipy` package behaves" point of view it seems good to go.

h-vetinari:
> Two timeouts for PyPy on `aarch64` it looks like.

Not even a normal 6h timeout, but two workers that stopped responding. Haven't had that one in a long time, just my luck that they appear when I declare things ready. 😅

In any case, I restarted and it shouldn't be an issue

xhochy (approved):
Looking good! Thanks @h-vetinari for bringing this here.

h-vetinari:
Seems we managed to drop about 6MB out of 19-20MB previously. Slightly less than hoped for, but better than nothing I'd say!

rgommers:
Great, thanks. I sent an FYI to the scipy-dev mailing list

PR #189, opened 2021-08-20
V1.5.4

thewchan:
Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -7,10 +7,10 @@ package:
 
 source:
   url: https://github.com/scipy/scipy/archive/v{{ version }}.tar.gz
-  sha256: 309f66c326e2a90b1ebb1a4ea75f3dd9c56276f392ec5a1b2fe986ea662bf69f
+  sha256: 4a453d5e5689de62e5d38edf40af3f17560bfd63c9c5bd228c18c1f99afa155b
 
 build:
-  number: 1
+  number: 0
   skip: true  # [py2k]
 
 requirements:
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "1.5.3" %}
+{% set version = "1.5.4" %}
 {% set ver_major_minor = ".".join(version.split(".")[:2]) %}
 
 package:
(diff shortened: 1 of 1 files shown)


thewchan:

h-vetinari:
Passing on py36 and failing elsewhere sounds like an incompatibility with newer numpy's (even though that shouldn't happen).

I'm not sure how much effort this is worth. It's possible to skip the tests, although cleaner would be to restrict the numpy version (not sure if `numpy <1.21` is sufficient, py36 is stuck on 1.19) - @thewchan, can your usecase handle a restriction to older numpy?

thewchan:
@h-vetinari 
Hmmm, the package I am trying to get on conda-forge says this in its `requirement.txt`:

```
numpy==1.19.1
```

Here's the link: [](https://github.com/shankarpandala/lazypredict)

h-vetinari:
@thewchan 
This requirements file is very likely over-specified (in fact, locked), particularly down to the patch version -- e.g. there's numpy 1.19.3, which is extremely likely to be compatible with 1.19.1 but probably has fewer bugs. I would start building the recipe without these explicit pins (down to the minor version if absolutely necessary) and see if the upstream test suite runs through.

conda recipes generally do not pin dependencies down that hard, because it becomes very hard to combine such packages with others in an environment. Lower and upper bounds make much more sense from the POV of the resolver and usage patterns.

Once you have a PR on staged-recipes, you can also ping (one of) the upstream maintainers if they want to collaborate - and whether they have opinions on the pinning.

thewchan:
In the event that upstream isn't responsive, is it possible to patch the requirement with looser pins?

dopplershift:
@thewchan No patching should be necessary, you're specifying the requirement in `meta.yaml` for that package.

thewchan:
I do already have a PR on staging, that's how I found out that it needed that version of Scipy, I'll try to relax the requirements for Scipy on that recipe and see if it works

https://github.com/conda-forge/staged-recipes/pull/15950

thewchan:
@dopplershift it seems that the 1.5.4 is hardcoded in the package

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=366025&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823

dopplershift:
In that case I would just avoid running `pip check`.

thewchan:
Got it, I'll try that @dopplershift

thewchan:
Resolved adding the other package without needing scipy 1.5.4, closing this PR. Thanks for everyone's help!

PR #193, opened 2021-10-26
Cross compile instead of emulating

isuruf:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -27,35 +27,35 @@ jobs:
       linux_aarch64_numpy1.18python3.7.____cpythonpython_implcpython:
         CONFIG: linux_aarch64_numpy1.18python3.7.____cpythonpython_implcpython
         UPLOAD_PACKAGES: 'True'
-        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-aarch64
+        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-comp7
       linux_aarch64_numpy1.18python3.8.____cpythonpython_implcpython:
         CONFIG: linux_aarch64_numpy1.18python3.8.____cpythonpython_implcpython
         UPLOAD_PACKAGES: 'True'
-        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-aarch64
+        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-comp7
       linux_aarch64_numpy1.19python3.7.____73_pypypython_implpypy:
         CONFIG: linux_aarch64_numpy1.19python3.7.____73_pypypython_implpypy
         UPLOAD_PACKAGES: 'True'
-        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-aarch64
+        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-comp7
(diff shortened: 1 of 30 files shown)


h-vetinari:
@mattip

It seems setuptools is expecting something from distutils that pypy might not be providing? Weird though that this only affects aarch64?
```
    File "/home/conda/feedstock_root/[...]/venv/lib/python3.7/site-packages/setuptools/command/egg_info.py", line 621, in _safe_data_files
      return build_py.get_data_files_without_manifest()
    File "/home/conda/feedstock_root/[...]/lib-python/3/distutils/cmd.py", line 103, in __getattr__
      raise AttributeError(attr)
```

mattip:
I recently fixed this for NumPy in numpy/numpy#20299 by pinning the setuptools version. A fix is already in setuptools in pypa/setuptools#2855, and I think it should be available soon as 58.5.3.

h-vetinari:
Yeah, I just saw this occur on another feedstock, and then found https://github.com/conda-forge/setuptools-feedstock/issues/208, as well as the issue you linked. The new setuptools version has already been packaged by conda-forge, only needs to get through the CDN now.

Sorry for the false alarm!

mattip:
On the one hand, projects with complicated workflows really should pin setuptools. But on the other-hand, if everyone uses old versions, setuptools will never discover it has broken someone's workflow. As suggested in the issue, if setuptools could somehow test its HEAD against more projects that would give an indication, but then it would slow down their CI significantly.

h-vetinari:
The pypy build on aarch hangs forever in `fft/tests/test_multithreading.py::test_mixed_threads_processes`, source [here](https://github.com/scipy/scipy/blob/v1.7.1/scipy/fft/tests/test_multithreading.py#L32-L43).

Any idea what could cause this @mattip?

(as before, the only build that was rerun is the last failing one; theoretically it could be that - like with the setuptools regression - other builds would be affected as well, but just haven't been been rerun due to previous success)

mariusvniekerk:
> The pypy build on aarch hangs forever in `fft/tests/test_multithreading.py::test_mixed_threads_processes`

We could potentially just skip that test for the pypy on aarch64 ?

mattip:
Skipping on pypy + aarch64 sounds reasonable. This is inside a qemu environment, right? It might be confusing the PyPy JIT. Although then I would expect other things to break as well. A first step may be to debug pypy + aarch64 in the pypy3.6-feedstock.

jakirkham:
It could also be oversubscription of the resources provided given it is using multiple threads (and/or processes)

h-vetinari:
I will take care of getting CI green (if need be by skipping tests) in the other two open PRs - first the version bump and then 3.10.

jakirkham:
Thanks all! 😀

PR #203, opened 2022-05-20
Scipy 1.8.1

h-vetinari:
Not sure what's up with the update bot


h-vetinari:
The time is not ripe for travis yet (maybe with meson for 1.9...), timing out in compilation on PPC, running into some infra error for aarch --> leaving this recipe as is...

PR #205, opened 2022-06-27
WIP / DNM: build with meson on unix

h-vetinari:
I don't expect this to work yet. While upstream support meson support is still maturing, we'll probably need to make the gfortan + clang hack work again. But such a fundamental build system change seemed like a good opportunity to try to see which work-arounds are still necessary.

* [ ] build against netlib
* [ ] make meson builds work
  * [ ] unix 
  * [ ] windows
* [ ] make BLAS variants work
* [ ] can we work around https://github.com/scipy/scipy/issues/16352?


h-vetinari:
Seems like conda's build isolation mechanism (with the placehold_placehold_...) is breaking ninja
```
ninja: error: stat(scipy/special/_specfun.cpython-310-x86_64-linux-gnu.so.p/_home_conda_feedstock_root_build_artifacts_scipy_1656354640264__h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place_lib_python3.10_site-packages_numpy_core_include_.._.._f2py_src_fortranobject.c.o): File name too long
```

Quite surprised about the appearance of windows-flavoured problems on unix... 🤔

h-vetinari:
> Seems like conda's build isolation mechanism (with the placehold_placehold_...) is breaking ninja

Interestingly, this does not seem to happen for aarch64/ppc64le, which then fail with:
```
  File "[...]/lib/python3.9/site-packages/scipy/_lib/_ccallback.py", line 1, in <module>
    from . import _ccallback_c
ImportError: cannot import name '_ccallback_c' from 'scipy._lib' ([...]/lib/python3.9/site-packages/scipy/_lib/__init__.py)
```

h-vetinari:
Does meson on windows only work with mingw? Surprised that MSVC seems to be detected as not working:
```
..\..\meson.build:1:0: ERROR: Compiler cl.exe can not compile programs.
```

All things considered, I'm probably going to need a bit of help on this PR @rgommers @eli-schwartz

h-vetinari:
Finally, for osx-arm, we might have to be more explicit about the fact that this should be cross-compiled:
```
../../meson.build:1:0: ERROR: Could not invoke sanity test executable: [Errno 86] Bad CPU type in executable
```

rgommers:
> While upstream support meson support is still maturing, we'll probably need to make the gfortan + clang hack work again

I don't think so, since that's not what we're using on Windows for wheels - there is no build system support for this. There's two options for 1.9.0:
1. Make the Windows build work with Mingw-w64
2. Still build the Windows packages with `numpy.distutils`.

> Does meson on windows only work with mingw? Surprised that MSVC seems to be detected as not working:

It works with any compiler toolchain that has coherent support for C/C++/Fortran. Which in practice is:
1. Mingw-w64 (with a bunch of fiddling needed to deal with forcing Mingw to treat `long double` as 64-bit rather than 80-bit)
2. (MSVC or ICC) + Intel Fortran

> Finally, for osx-arm, we might have to be more explicit about the fact that this should be cross-compiled:

Ah, that should work but I have not dealt with the specifics of cross-compiling yet (also not for the macOS arm64 wheel builds). For documentation, see https://mesonbuild.com/Cross-compilation.html. I'm guessing we'll have to figure out what the cross file for conda-forge builds should look like.

eli-schwartz:
From the failing logs:

```
..\..\meson.build:1:0: ERROR: Compiler cl.exe can not compile programs.

A full log can be found at C:\bld\scipy_1656354772652\work\.mesonpy-902z_jt4\build\meson-logs\meson-log.txt
```

This log file contains debug information about the failed state including the command lines, input files, and console output for compiler checks.

Not sure how easy it is to get at that underneath the mesonpy build backend (it might clean up the build directory automatically? Don't know).

h-vetinari:
> There's two options for 1.9.0:
>  1. Make the Windows build work with Mingw-w64

I'll need to defer to @isuruf if he considers that feasible here (i.e. why he chose MSVC as the main compiler when adding windows support). 

>  2. Still build the Windows packages with `numpy.distutils`.

That's obviously possible for 1.9, but not a permanent solution AFAIU - isn't non-meson support planned to be removed in 1.10?

> > Does meson on windows only work with mingw? Surprised that MSVC seems to be detected as not working:
> 
> It works with any compiler toolchain that has coherent support for C/C++/Fortran. Which in practice is:

Regarding compilers, it's perhaps worth considering if we can move everything to the LLVM stack on windows? We'd probably have to update the (classic-)flang builds a bit, but in principle that should be an option?

h-vetinari:
Thanks for the input @eli-schwartz! I think Ralf pinpointed the issue regarding the consistent toolchain, or rather the lack thereof. Windows builds here currently use an unholy combination of MSVC+gfortran, which presumably meson doesn't recognize (and arguably it shouldn't).

rgommers:
> (i.e. why he chose MSVC as the main compiler when adding windows support).

It was the only option and matching the way SciPy wheels were built at the time.

> That's obviously possible for 1.9, but not a permanent solution AFAIU - isn't non-meson support planned to be removed in 1.10?

Yes, but that's fine. That gives us 6 months to solve the problem. And either way, switching the build away from `distutils` should be done in a separate PR because it's a very much nontrivial change. So let's not mix it with changes for the other supported platforms (where I would not expect significant trouble).

rgommers:
@h-vetinari note that there's some lingering problems with wheels builds, I won't be sure that this all works as advertised until https://github.com/MacPython/scipy-wheels/pull/171 is complete.

h-vetinari:
> @h-vetinari note that there's some lingering problems with wheels builds

Normally I'd say wheel builds don't concern us directly, but now I don't know if they necessarily get built as an intermediate step in the meson builds?

In any case, I opened #206 for a distutils-based test of the rc, which looks fine except for highs on windows.

rgommers:
> Normally I'd say wheel builds don't concern us directly, but now I don't know if they necessarily get built as an intermediate step in the meson builds?

They're not necessary technically (with `meson` or with `distutils`), but we're already using them: https://github.com/conda-forge/scipy-feedstock/blob/main/recipe/build.sh#L9-L13 as it's an easier interface.

That's not quite what I meant though. Before we have done a single release with wheels, you may run into issues here that we have not yet found (example: I just noticed that the main `LICENSE.txt` file will be missing, which will also affect this PR).

h-vetinari:
> They're not necessary technically (with `meson` or with `distutils`), but we're already using them: https://github.com/conda-forge/scipy-feedstock/blob/main/recipe/build.sh#L9-L13 as it's an easier interface.

Sure, but we don't have to do anything but install them and can discard them right away (no auditwheel, etc.)



> That's not quite what I meant though. Before we have done a single release with wheels, you may run into issues here that we have not yet found (example: I just noticed that the main `LICENSE.txt` file will be missing, which will also affect this PR).

Understood. Similarly for the HiGHS compilation error on windows I presume. Do you want me to raise an issue?

rgommers:
> Similarly for the HiGHS compilation error on windows I presume. Do you want me to raise an issue?

No that's fine, I think that belongs on this PR, the error is expected.

h-vetinari:
> > Similarly for the HiGHS compilation error on windows I presume. Do you want me to raise an issue?
> 
> No that's fine, I think that belongs on this PR, the error is expected.

I don't understand. It happens in #206 (not this PR), and that one's using the non-meson build path, so I'd expect a green test suite everywhere, but get a build error in the new submodule (haven't found a similar issue or PR tackling that on the upstream repo).

rgommers:
> I don't understand. It happens in #206 (not this PR), and that one's using the non-meson build path, so I'd expect a green test suite everywhere, but get a build error in the new submodule (haven't found a similar issue or PR tackling that on the upstream repo).

Ah I missed that part. In that case, yes please.

h-vetinari:
> In that case, yes please.

Done

rgommers:
It's probably time to switch anything but the Windows builds to Meson now. SciPy released 1.9.3 and 1.10.0 wheels built with Meson, things should work. And the `--config-settings` UX change in `meson-python` is present now and can be used to build against the desired BLAS/LAPACK flavor for conda-forge.

WDYT @h-vetinari?

h-vetinari:
> WDYT @h-vetinari?

SGTM!

I wanted to get 1.10 out the door (and the CI was green...), before going down the rabbit hole of how much changes will _actually_ be necessary.

rgommers:
I'd expect very little, leaving aside the Windows issues:
- ~switch back to `pip install .`~ use `python -m build` in `build.sh`
- use the `--use-g77-abi` CLI flag instead of the env var
- specify `blas` and `lapack` through CLI options

So (untested) something like:
```
# -wnx flags mean: --wheel --no-isolation --skip-dependency-check
python -m build -w -n -x  -Csetup-args=-Dblas=blas -Csetup-args=-Dlapack=lapack -Csetup-args=-Duse-g77-abi=true
pip install dist/scipy*.whl
```

(yes this is a bit ugly, and can't use `pip` because it doesn't accept multiple times the same key for `--config-settings` - see https://github.com/mesonbuild/meson-python/issues/230#issuecomment-1326888653)

h-vetinari:
> So (untested) something like:
> 
> ```
> # -wnx flags mean: --wheel --no-isolation --skip-dependency-check
> python -m build -w -n -x  -Csetup-args=-Dblas=blas -Csetup-args=-Dlapack=lapack -Csetup-args=-Duse-g77-abi=true
> pip install dist/scipy*.whl
> ```

Good guess, this worked on unix in native compilation.

I guess it's expected that cross-compilation fails?
```
ERROR Backend subprocess exited when trying to invoke build_wheel
```

Happy to dig into this of course, but if you have some references handy of where I need to tackle this, that would be very nice. :)

rgommers:
Relevant part of the build log:
```
+CXX=$BUILD_PREFIX/bin/aarch64-conda-linux-gnu-c++
-CXX=$PREFIX/bin/aarch64-conda-linux-gnu-c++
-DEBUG_CXXFLAGS=-fvisibility-inlines-hidden -fmessage-length=0 -ftree-vectorize -fPIC -fstack-protector-all -fno-plt -Og -g -Wall -Wextra -fvar-tracking-assignments -pipe -isystem $PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/scipy-1.10.0 -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix
+DEBUG_CXXFLAGS=-fvisibility-inlines-hidden -fmessage-length=0 -ftree-vectorize -fPIC -fstack-protector-all -fno-plt -Og -g -Wall -Wextra -fvar-tracking-assignments -pipe -isystem $PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/scipy-1.10.0 -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix -fvisibility-inlines-hidden -fmessage-length=0 -ftree-vectorize -fPIC -fstack-protector-all -fno-plt -Og -g -Wall -Wextra -fvar-tracking-assignments -pipe -isystem $PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/scipy-1.10.0 -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix
+GXX=$BUILD_PREFIX/bin/aarch64-conda-linux-gnu-g++
-GXX=$PREFIX/bin/aarch64-conda-linux-gnu-g++
Setting up cross-python
Finished setting up cross-python
+ /home/conda/feedstock_root/build_artifacts/scipy_1674797616497/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place/bin/python -m build -w -n -x -Csetup-args=-Dblas=blas -Csetup-args=-Dlapack=lapack -Csetup-args=-Duse-g77-abi=true
* Building wheel...
+ meson setup --prefix=$PREFIX $SRC_DIR $SRC_DIR/.mesonpy-8brkxeub/build --native-file=$SRC_DIR/.mesonpy-native-file.ini -Ddebug=false -Doptimization=2 -Dblas=blas -Dlapack=lapack -Duse-g77-abi=true
```

We're not telling `meson` that we're cross-compiling here. That needs a cross file, see https://mesonbuild.com/Cross-compilation.html. I think conda-forge has a way of doing that already for Meson. General docs are at https://conda-forge.org/docs/maintainer/knowledge_base.html#cross-compilation, but they don't tell us enough. https://conda-forge.org/blog/posts/2020-10-29-macos-arm64/ says to add `meson ${MESON_ARGS} builddir/`, and that should fix things. I hope that's still unchanged.

Note: if that works, then it's likely that there will be a problem with `libnpymath`, I've seen that before when cross-compiling. Not sure though, let's figure that out if and when it happens.

rgommers:
To expand on the above, we still have a problem of glueing the commands together:
- native build: `python -m build ...` produces a wheel in one go
- cross-build: conda-forge wants us to run `meson ${MESON_ARGS} builddir/`, _however_ that may conflict with `meson-python` trying to re-run the configure stage of the build.

So first try I'd say:
```
meson ${MESON_ARGS} builddir
python -m build ...<existing-args-here> -Cbuilddir=builddir
```
xref https://meson-python.readthedocs.io/en/latest/usage/build-options.html#builddir for reusing an existing build directory.

rgommers:
If you run that once and there's a problem, I can debug this locally on macOS arm64. I just need to know what `MESON_ARGS` contains.

h-vetinari:
OK, a few strange things happening. On linux-aarch (similarly for ppc), we apparently build a wheel successfully:
```
Successfully built scipy-1.10.0-cp310-cp310-linux_aarch64.whl
```
but all the libs have been compiled for the build env, and not the host:
```
[35/1241] scipy/linalg/_fblas.cpython-310-x86_64-linux-gnu.so
[36/1241] scipy/linalg/_flapack.cpython-310-x86_64-linux-gnu.so
[37/1241] scipy/linalg/_flinalg.cpython-310-x86_64-linux-gnu.so
[38/1241] scipy/linalg/_interpolative.cpython-310-x86_64-linux-gnu.so
```

Predictably this fails upon import. So far I used the conda-forge default as in the docs you linked (so `${MESON_ARGS}` and the auto-generated cross-build-definition file), but I guess we'll have to dig where that goes wrong.

On osx-arm, we get `Found Pkg-config: NO`, even though pkg-config is in both the build & the host env. Consequently, blas detection fails and we fall on our nose.

recipe/build.sh
```diff
@@ -1,9 +1,15 @@
...
 
-# Use the G77 ABI wrapper everywhere so that the underlying blas implementation
-# can have a G77 ABI (currently only MKL)
-export SCIPY_USE_G77_ABI_WRAPPER=1
+# need to run meson first for cross-compilation case
+meson ${MESON_ARGS}
+    -Dblas=blas \
+    -Dlapack=lapack \
+    -Duse-g77-abi=true \
+    builddir || (cat builddir/meson-logs/meson-log.txt && exit 1)
```
rgommers, line 11: `builddir` is in the wrong place here. Can you remove it from this line, and make the first line:
```
meson setup builddir ${MESON_ARGS} \
```
(note that `\` was missing on line 7, which explains some failures.

rgommers:
The macOS x86-64 build shows:
```
mkdir builddir
meson --buildtype release --prefix=/Users/runner/miniforge3/conda-bld/scipy_1675668478043/_h_env_placehold_placehold_<etc> -Dlibdir=lib
ERROR: Must specify at least one directory name.
```

`MESON_ARGS` doesn't look all that useful to me, but it can't hurt either I guess.

recipe/build.sh
```diff
@@ -1,9 +1,15 @@
...
+set -ex
 
-set -x
+mkdir builddir
 
-# Use the G77 ABI wrapper everywhere so that the underlying blas implementation
-# can have a G77 ABI (currently only MKL)
-export SCIPY_USE_G77_ABI_WRAPPER=1
+# need to run meson first for cross-compilation case
+meson ${MESON_ARGS}
```
h-vetinari, line 7:
```suggestion
meson ${MESON_ARGS} \
```

recipe/build.sh
h-vetinari: OK, I was going to come back to this PR today anyway 👍

eli-schwartz:
```
meson <args> builddir

meson setup <args> builddir

meson setup builddir <args>
```

Currently, this uses the first line, you're suggesting that only the third line is "correct". This is technically not true -- only the second line is "correct". It is in the `meson setup --help` text.

It's correct that neglecting the word "setup" is wrong -- meson warns about it now.

The `-D` words are options/option-arguments to "setup". The "builddir" word is an operand. Naming conventions come from the [POSIX / ISO 9945 utility conventions](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html). Some people are more familiar with:
- options/option-arguments as, variously, "parameters", "arguments",  or "flags"
-  operands as "arguments" or "parameters".

Python's argparse, which Meson uses, calls:
- options -> "flags", "options", or "optional arguments"
- option-arguments -> "options", or "optional arguments"
- operands -> "positional arguments"

The POSIX utility conventions section 2, Utility Syntax Guidelines, state that options should always precede operands. GNU-style argument parsing differs from this -- it supports argument permutation, and whenever it finds an operand it shifts it toward the end and returns options/option-arguments first. argparse supports that, but not by default -- and Meson doesn't use `parse_intermixed_args()`, so we do not get the GNU style. But argparse *does* permit specifying *all* operands before options/option-arguments.

So returning to the 3 example Meson invocations -- argparse fundamentally supports both lines two and three, and treats them identically. But, the --help text will describe the use of line two.

...

tl;dr what to do?

Honestly, do whatever you want to do. But if it's a matter of technical correctness, leave "builddir" at the end. :D

h-vetinari:
Well, it seems no cross-file is actually being written (and `MESON_ARGS` populated) except for osx-arm...

recipe/build.sh
h-vetinari: Thanks for the insights. For reference, https://mesonbuild.com/Cross-compilation.html has a call with the options after `builddir/`, and one without `setup`, so it's perhaps not yet homogeneously used in the docs.

eli-schwartz: Ideally `setup` would be homogeneously used in the docs, but I could have possibly missed some when doing mass rewrites. :) It is of course totally wrong, because Meson will actively warn you if you follow the docs...

If you would like to submit a PR to fix the documentation wherever you catch stale references to a missing `setup`, that would be great. Otherwise I'll take a look when I get a chance.

(Again, the order of options simply doesn't matter, but there's a pedant's approach to correctness that suggests one way over the other, and no one will ever accuse me of not enjoying an excuse to pontificate about POSIX.)

isuruf:
gcc activation scripts don't set MESON_ARGS yet unlike the clang activation scripts. PRs are welcome.

h-vetinari:
Seems pkgconfig and cmake aren't being picked up by meson in cross-compilation mode...
```
Pkg-config binary for 1 is cached as not found.
Dependency lookup for blas with method 'pkgconfig' failed: Pkg-config binary for machine 1 not found. Giving up.
CMake binary for 1 is not cached
CMake binary missing from cross or native file, or env var undefined.
Default target is not allowed for cross use
Found CMake: NO
```



> gcc activation scripts don't set MESON_ARGS yet unlike the clang activation scripts. PRs are welcome.

Once I have a better grasp what's necessary I'll do that.

eli-schwartz:
> Seems pkgconfig and cmake aren't being picked up by meson in cross-compilation mode...

Meson cannot know which pkgconfig is the correct pkgconfig in cross compilation mode. `pkg-config` from PATH is most likely incorrect, because it will refer to non-cross-compiled versions of libraries and headers.

So you need to add this to your cross file:
```ini
[binaries]
pkgconfig = '/path/to/{triplet}-pkg-config'
```

which is often a wrapper shell script that sets $PKG_CONFIG_LIBDIR and a couple other variables, then runs the regular pkg-config. But it might also be a pkg-config that was compiled to target the cross environment.

cmake is a similar story, it's frequently the same cmake binary as the native one but with the relevant `$CMAKE_*` variables set to use a custom prefix/sysroot and override the module search path to find cross-built *-config.cmake files instead of native ones.

recipe/build.sh
rgommers: Good point, I inferred the needed order of arguments from the error message, but that was caused by a backslash missing and not by the position of `builddir`.

h-vetinari:
So, after some trial and error, I'm getting meson to pick up the cross-file (+pkgconfig & cmake) correctly,
```
Build machine cpu family: x86_64
Build machine cpu: x86_64
Host machine cpu family: aarch64
Host machine cpu: aarch64
Target machine cpu family: aarch64
Target machine cpu: aarch64
```
but it still generates `*.cpython-38-x86_64-linux-gnu.so`, i.e. for the wrong architecture.

Could it be that our cross-python does not interact well with meson? 🤔

rgommers:
> but it still generates `*.cpython-38-x86_64-linux-gnu.so`, i.e. for the wrong architecture.
> 
> Could it be that our cross-python does not interact well with meson?

So build and host machine configuration looks fine:
```
Build machine cpu family: x86_64
Build machine cpu: x86_64
Host machine cpu family: aarch64
Host machine cpu: aarch64
```
but then it probably gets messed up by this reconfigure step:
```
* Building wheel...
+ meson setup --prefix=$PREFIX $SRC_DIR $SRC_DIR/builddir --native-file=$SRC_DIR/.mesonpy-native-file.ini -Ddebug=false -Doptimization=2
Directory already configured.

Just run your build command (e.g. ninja) and Meson will regenerate as necessary.
If ninja fails, run "ninja reconfigure" or "meson setup --reconfigure"
to force Meson to regenerate.

If build failures persist, run "meson setup --wipe" to rebuild from scratch
using the same options as passed when configuring the build.
To change option values, run "meson configure" instead.
+ meson compile
INFO: autodetecting backend as ninja
INFO: calculating backend command to run: $BUILD_PREFIX/bin/ninja
[1/1629] Generating scipy/generate-version with a custom command
...
[11/1629] Compiling C object scipy/_lib/_test_ccallback.cpython-310-x86_64-linux-gnu.so.p/src__test_ccallback.c.o
```

I'm not exactly sure if that is what is happening, but it's either that or a bug in `meson` or `meson-python` which gives us the wrong `x86_64` in the extension module names.

I don't know anything about `cross-python` though. Looking at https://github.com/conda-forge/cross-python-feedstock/blob/main/recipe/activate-cross-python.sh, it uses [crossenv](https://pypi.org/project/crossenv/) and then has some build backend or Python interpreter specific changes. `crossenv` hasn't been tried before for cross-compiling SciPy AFAIK; regular native and host interpreters did work before.

h-vetinari:
So now, definitely without running reconfigure, we still end up compiling for x64. CC @eli-schwartz 

```
+ meson setup --prefix=/home/conda/feedstock_root/build_artifacts/scipy_1676716693801/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place --libdir=lib --cross-file=/home/conda/feedstock_root/build_artifacts/scipy_1676716693801/_build_env/meson_cross_file.txt -Dblas=blas -Dlapack=lapack -Duse-g77-abi=true builddir
The Meson build system
[...]
Build machine cpu family: x86_64
Build machine cpu: x86_64
Host machine cpu family: aarch64
Host machine cpu: aarch64
[...]
Build targets in project: 253

SciPy 1.10.0

  User defined options
    Cross files: $BUILD_PREFIX/meson_cross_file.txt
    libdir     : lib
    prefix     : $PREFIX
    blas       : blas
    lapack     : lapack
    use-g77-abi: true

Found ninja-1.11.0 at $BUILD_PREFIX/bin/ninja
WARNING: custom_target '_stats_gen_pyx' has more than one output! Using the first one.
[...]
WARNING: custom_target 'cython_linalg' has more than one output! Using the first one.
+ meson install -C builddir
ninja: Entering directory `$SRC_DIR/builddir'
[1/1629] Generating scipy/generate-version with a custom command
[...]
[14/1629] Linking target scipy/_lib/_test_ccallback.cpython-38-x86_64-linux-gnu.so
```

eli-schwartz:
```
Program cython found: YES ($BUILD_PREFIX/bin/cython)
Program python3 found: YES ($BUILD_PREFIX/bin/python3.11)
Found pkg-config: $BUILD_PREFIX/bin/pkg-config (0.29.2)
Program pythran found: YES ($PREFIX/bin/pythran)
```

Is $BUILD_PREFIX/bin/python3.11 an aarch64 python? I think you need that because it must be able to `import numpy; print(numpy.get_include())`, and if it points to the wrong python then you also end up compiling against the wrong numpy.

If not for the need to import numpy, it would be possible to use an x86_64 python using `$_PYTHON_SYSCONFIGDATA_NAME` to have it report suitable information for a cross python.

...

If that python3.11 is an aarch64 binary but it's still reporting mysterious x86_64 information, then that's probably a Meson bug.

rgommers:
> Is $BUILD_PREFIX/bin/python3.11 an aarch64 python? I think you need that because it must be able to `import numpy; print(numpy.get_include())`, and if it points to the wrong python then you also end up compiling against the wrong numpy.
> 
> If not for the need to import numpy, it would be possible to use an x86_64 python using `$_PYTHON_SYSCONFIGDATA_NAME` to have it report suitable information for a cross python.

Good catch. I believe it's not. And if it were, then running it to obtain the `numpy`, `pybind11` and `pythran` include directories would not work.

So there's two things here: (1) a missing piece in SciPy's cross-compile puzzle, and (2) how to select the right Python interpreter for the conda-forge builds.

For (2), the linux-aarch64 build log shows this as the contents of the cross file:

```ini
cmake = '$BUILD_PREFIX/bin/cmake'
pkgconfig = '$BUILD_PREFIX/bin/pkg-config'

[host_machine]
system = 'linux'
cpu = 'aarch64'
cpu_family = 'aarch64'
endian = 'little'

[properties]
pkg_config_libdir = '$PREFIX/lib'
```

And then the Python interpreter is found in the top-level `meson.build` file through:
```meson
py_mod = import('python')
py3 = py_mod.find_installation(pure: false)                            
py3_dep = py3.dependency()
```

I think the `meson.build` contents are fine; it's not clear to me if Meson uses  `pkg-config` to look for `python` and that is going wrong (maybe because `pkg_config_libdir` is not correct or `python.pc` is missing?), or if Meson uses its builtin `python` detection capabilities first and then picks the wrong `python`.

For _(1) a missing piece in SciPy's cross-compile puzzle_ - that, unfortunately, is still in my backlog. There's a hack we may have to apply here to hardcode the paths rather than try to run the host python to obtain the include dirs. The structural solution is https://github.com/mesonbuild/meson/issues/9598#issuecomment-1193322394, which depends on the Python refactor that you're trying to land right now @eli-schwartz.

h-vetinari:
Thanks both for your inputs @eli-schwartz & @rgommers!

> For (2), the linux-aarch64 build log shows this as the contents of the cross file:

I think this isn't news to anyone, but just to be sure, this was me hacking together the cross-file based on what looked relevant in the docs. It's entirely possible that I messed up (either included too little or too much, or set the wrong values).

rgommers:
> The structural solution is [mesonbuild/meson#9598 (comment)](https://github.com/mesonbuild/meson/issues/9598#issuecomment-1193322394), which depends on the Python refactor that you're trying to land right now @eli-schwartz.

Or maybe that is closer than I realized - I missed that `find_installation` has a `modules` argument already that we should be using: https://mesonbuild.com/Release-notes-for-0-51-0.html#new-modules-kwarg-for-pythonfind_installation. I'll revisit this.

h-vetinari:
> > The structural solution is [mesonbuild/meson#9598 (comment)](https://github.com/mesonbuild/meson/issues/9598#issuecomment-1193322394), which depends on the Python refactor that you're trying to land right now @eli-schwartz.
> 
> Or maybe that is closer than I realized - I missed that `find_installation` has a `modules` argument already that we should be using: https://mesonbuild.com/Release-notes-for-0-51-0.html#new-modules-kwarg-for-pythonfind_installation. I'll revisit this.



@rgommers, trying to follow the various different threads here (https://github.com/mesonbuild/meson/pull/11392 being merged but not released in meson, your comments in https://github.com/mesonbuild/meson/issues/9598 & https://github.com/scipy/scipy/pull/18006 merged but unreleased in scipy). For now I'm giving it a shot to backport the scipy PR. It's also conceivable to build a dev version of meson under a different "label" (properly published but not installable from conda-forge without specifying an extra channel), but I'd have to see if I can join the feedstock or convince people to do that.

h-vetinari:
I gave it a try to do the whole thing without cross-compilation in https://github.com/conda-forge/scipy-feedstock/pull/229. Perhaps unsurprisingly, this works fine on linux-aarch/ppc. This would allow us to get everything but windows and osx-arm switched to meson, however, we'll still need to fix cross-compilation for osx-arm anyway, and we'd lose the ability to run the test suite on the emulated builds, because emulated build + emulated test suite runs into the 6h timeout (whereas cross-compiled build + emulated test suite is feasible).

eli-schwartz:
How does the emulation work?

h-vetinari:
> How does the emulation work?

Simulating that the build running on aarch64 (resp. ppc64le), by going through QEMU (which is running on a x64 agent). The emulation adds an overhead of ~4-6x to the runtime, compared to native (resp. cross) compilation.

eli-schwartz:
So qemu user-mode emulation to run the build scripts, rather than qemu to boot a VM?

h-vetinari:
> So qemu user-mode emulation to run the build scripts, rather than qemu to boot a VM?

Yes.

rgommers:
> For now I'm giving it a shot to backport the scipy PR.

That shouldn't actually help directly for cross-compilation, only for dealing with Docker and venv-in-source-tree issues. I am about to write a second PR on top that allows bypassing running any of that code by hardcoding the paths to numpy/pybind11/pythran.

isuruf:
Can meson build for a python different to the one meson is run on? Is there a way to pass the python interpreter to use for the package being built?

rgommers:
> Can meson build for a python different to the one meson is run on? Is there a way to pass the python interpreter to use for the package being built?

Yes, it has good support for cross compilation: https://mesonbuild.com/Cross-compilation.html#cross-compilation. This needs a "cross file" to identify the relevant binaries: (example: https://github.com/scipy/scipy/blob/maintenance/1.9.x/tools/wheels/cross_arm64.txt). I think you can explicitly point to the host Python binary, or rely on pkg-config for that.

I'd say we have issues now because SciPy tries to run the host Python. `crossenv` is kinda a nice hack, because it installs a "host Python" that's really a native Python with tweaks. Ideally we'd get rid of the need to run the host interpreter at all, at which point regular cross-compiling should work (modulo issues with `sysconfig` etc. - I know it's not trivial).

h-vetinari:
> The trick is specifying that it's a cross build _and defining the cross python_ as a qemu-user wrapped binary.

How would that look like in the crossfile?

h-vetinari:
So I backported scipy/scipy#18034, but it's worth noting that Isuru managed to get the compilation to run correctly by using `$PYTHON $(which meson)`. However, I'm not sure we'd be able to tell `$PYTHON -m build -w -n -x -Cbuilddir=builddir` to use the right meson here?

In any case, using that patch with the vanilla `$PYTHON -m build -w -n -x -Cbuilddir=builddir`, we get the right ABI, but are then failing in a meson check, presumably because it's picking up the build and not the host version.

```
  File "/home/conda/feedstock_root/build_artifacts/scipy_1677237403228/_build_env/venv/lib/python3.11/site-packages/mesonpy/__init__.py", line 325, in _stable_abi
    raise ValueError(
ValueError: Extension module 'scipy/_lib/_ccallback_c.cpython-311-aarch64-linux-gnu.so' not compatible with Python interpreter. Filename suffix '.cpython-311-aarch64-linux-gnu.so' not in {'.so', '.abi3.so', '.cpython-311-x86_64-linux-gnu.so'}.
```

h-vetinari:
Interestingly, osx-arm passed... 🤔 

Not sure yet what the difference is between that and the cross-compilation on linux.

rgommers:
> However, I'm not sure we'd be able to tell `$PYTHON -m build -w -n -x -Cbuilddir=builddir` to use the right meson here?

`meson-python` uses a native file to explicitly select the Python interpreter to use. Which in turn should then be using the correct `meson` IIRC. I'm a little fuzzy though on why there's more than one version of `meson` installed to begin with. Does by default every build dependency get installed for both build and host?

> Extension module 'scipy/_lib/_ccallback_c.cpython-311-aarch64-linux-gnu.so' not compatible

xref https://github.com/mesonbuild/meson-python/issues/321

h-vetinari:
> I'm a little fuzzy though on why there's more than one version of `meson` installed to begin with. Does by default every build dependency get installed for both build and host?

No, only as specified in build/host envs. It's possible we need meson _only_ in build...

rgommers:
> It's possible we need meson _only_ in build...

Yes indeed, that should be the case when we're cross-compiling. Same for other build tools and code generators, like `ninja`, `meson-python`, `cython`. And probably `pythran` too (I'm double-checking that one).

h-vetinari:
So far, cython was in both build and host even when cross-compiling. I'll try moving meson & ninja to build-only first. Probably will take a bit of iteration, but have to drop off here for now.

h-vetinari:
Mainly I'm waiting for the patch for https://github.com/mesonbuild/meson-python/issues/321 to land, so we can backport it to conda-forge's meson. I think with that we'd solve things (as of https://github.com/conda-forge/scipy-feedstock/pull/205/commits/c9afb89c183b2fc321f5cb6fc86b8547911e621c).

In the meantime, trying https://github.com/scipy/scipy/pull/18034 + specifying the include directories, I get
```
scipy/meson.build:81:0: ERROR: Include dir $PREFIX/lib/python3.9/site-packages/numpy/core/include does not exist.
```
even if those definitely exist.

rgommers:
> `ERROR: Include dir $PREFIX`

The `PREFIX` variable here doesn't look like it's substituted for the actual prefix. The string shown by Meson should be an absolute path.

h-vetinari:
Seems pkgconfig and cmake aren't being picked up by meson in cross-compilation mode...
```
Pkg-config binary for 1 is cached as not found.
Dependency lookup for blas with method 'pkgconfig' failed: Pkg-config binary for machine 1 not found. Giving up.
CMake binary for 1 is not cached
CMake binary missing from cross or native file, or env var undefined.
Default target is not allowed for cross use
Found CMake: NO
```



> gcc activation scripts don't set MESON_ARGS yet unlike the clang activation scripts. PRs are welcome.

Once I have a better grasp what's necessary I'll do that.

> The `PREFIX` variable here doesn't look like it's substituted for the actual prefix. The string shown by Meson should be an absolute path.

That was my suspicion as well, but with the last debug commit, I'm using the same vanilla sed that worked before, and I'm 99% certain the content in the crossfile is correct, i.e. an absolute path. (It should work at least for python 3.9, because I had to choose one python version to replace SP_DIR with)

PS. It's confusing that `cat` will sometimes replace expanded path variables back to environment variables.

isuruf:
> The PREFIX variable here doesn't look like it's substituted for the actual prefix. The string shown by Meson should be an absolute path.

conda-build captures stdout and replaces the values with $PREFIX. It makes the logs much easier to read, but might be confusing to new users.

h-vetinari:
FYI: I opened up a cleaned up version of this PR in #231, with the intention of merging it in the near future.

@isuruf, after rebasing out the later-reverted-anyway bf4313407aa6c5bc21fcae172ae0a63c7c711ac2, the modifications you made in aa19c0087379314a22ae921a0f8d6e8a130d23e9 & c4a44be9bfb98bc8915061dec86723611eded6a3 only needed to be done once each, rather than twice (because there's now one less meson-invocation). To my mind, the adaptation to this fact does not impact the faithfulness of the rebase to your changes, but please let me know if you feel otherwise.

PR #221, opened 2022-12-14
scipy 1.10.0

h-vetinari:
scipy 1.10rc1 was released, let's try building it!


h-vetinari:
@rgommers @tylerjereddy
rc1 is looking pretty good in conda-forge, aside perhaps from pypy on win (which is not exactly a surprise; we never published 1.9 for pypy, for example, see #208). 👍 

Note that this is still using the distutils path. I'm planning to immerse myself in the meson side of things in the new year.

rgommers:
That sounds great, thanks for checking @h-vetinari!

PR #229, opened 2023-02-23
WIP: meson without cross-compilation

h-vetinari:
Respin of #205 without requiring cross-compilation for linux-aarch/ppc. We still need to fix cross-compilation (at the very least for osx-arm), but perhaps this way we can achieve some intermediate goals on the way there.


rgommers:
Ah, with using QEMU things work, but the build time is a little excessive. Good data point; less sure that we should actually merge this unless/until we really have an urgent need. Probably better to push a bit harder on cross-compilation?

h-vetinari:
Closing as obsolete

PR #243, opened 2023-06-28
SciPy 1.11.1

h-vetinari:
A quick release to fix a licensing [issue](https://github.com/scipy/scipy/issues/18765) (scipy actually yanked 1.11.0; not sure if we want to mark the builds as broken)

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "1.11.0" %}
+{% set version = "1.11.1" %}
 
 package:
   name: scipy-split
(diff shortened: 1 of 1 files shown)


h-vetinari:
From a quick calculation based on [this](https://anaconda.org/conda-forge/scipy/files), we've already had >50k downloads.

rgommers:
When you mark a build as broken, will it still be installable when the exact version is contained in, for example, a lock file produced by `conda-lock`?

h-vetinari:
No, the build is then truly unavailable and will break reproducibility of any environment that got created in the meantime. This works differently than the yanking mechanism on PyPI.

There is a way to reproduce the environment, but I guess it's not so widely known, which is to add a special channel `-c conda-forge/label/broken`, then the environment can be reproduced.

But it's a big hammer, and I honestly doubt it's worth it

rgommers:
> But it's a big hammer, and I honestly doubt it's worth it

Agreed I think (unless there's a conda-forge policy that says the opposite) - it was yanked because of one file that shouldn't been removed, and that file wasn't even used in the build.

PR #246, opened 2023-09-03
WIP: meson on windows (clang-cl.exe)

h-vetinari:
I want to say "it's finally happening", but it'll probably still be a decent chunk of work to get this off the ground...

At least we got https://github.com/conda-forge/flang-feedstock/pull/28 in now. However, the binary is still named `flang-new`, not sure if meson can deal with this yet (upstream naming [discussion](https://discourse.llvm.org/t/proposal-rename-flang-new-to-flang/69462/33) in LLVM has more details; last blocker for rename has an [RFC](https://discourse.llvm.org/t/rfc-enabling-the-hlfir-lowering-by-default/72778) already...).

Then there's the question of mixing msvc, clang & flang, etc. etc...


h-vetinari:
> However, the binary is still named `flang-new`, not sure if meson can deal with this yet (upstream naming [discussion](https://discourse.llvm.org/t/proposal-rename-flang-new-to-flang/69462/33) in LLVM has more details; last blocker for rename has an [RFC](https://discourse.llvm.org/t/rfc-enabling-the-hlfir-lowering-by-default/72778) already...).

Yup, seems that meson doesn't now `flang-new`:
```
..\meson.build:82:0: ERROR: Unknown compiler(s): [['ifort'], ['gfortran'], ['flang'], ['pgfortran'], ['g95']]
The following exception(s) were encountered:
Running `ifort --version` gave "[WinError 2] The system cannot find the file specified"
Running `ifort -V` gave "[WinError 2] The system cannot find the file specified"
Running `gfortran --version` gave "[WinError 2] The system cannot find the file specified"
Running `gfortran -V` gave "[WinError 2] The system cannot find the file specified"
Running `flang --version` gave "[WinError 2] The system cannot find the file specified"
Running `flang -V` gave "[WinError 2] The system cannot find the file specified"
[...]
```

@isuruf [wanted](https://github.com/conda-forge/flang-feedstock/pull/28#issuecomment-1702057278) to keep the name as-is for now. Given the indications that upstream flang will do the rename to `flang` soon-ish, I'm not sure that meson will want to wire up `flang-new`? Is there a way to hack around that @eli-schwartz? Perhaps we could do this on the meson feedstock (rather than upstream)?

Otherwise, I'd be fine with renaming `flang-new` -> `flang` within conda-forge. It's not like the "classic" flang and llvm-based flang could ever be installed into the same environment, so I don't see a big problem there.

eli-schwartz:
`FC=flang-new meson setup builddir` will override the default iterative probing of various possible compiler binaries most of which you don't want. Exactly like `$CC` or `$CXX`.

This is *necessary* if you have multiple compilers installed and you need to use a specific one and ensure the wrong one isn't used. ;) In this case it will just let you detect a compiler you might not otherwise have found at all.

This does also assume that meson knows how to identify it based on the `--version` output, so you'll need to try it and see. 

I would prefer to avoid adding probes for an experimental-only name,  but if it still doesn't work after overriding the compiler name then I'd be happy to help get that working.

h-vetinari:
Thanks for the super quick response! 🙏
I'll try this tomorrow.

h-vetinari:
This is gonna be a whack-a-mole for a while I think 🙈 
```
..\meson.build:82:0: ERROR: Unable to detect linker for compiler `flang-new -Wl,--version`
stdout: LINK : warning LNK4044: unrecognized option '/-version'; ignored
LINK : warning LNK4001: no object files specified; libraries used
LINK : warning LNK4068: /MACHINE not specified; defaulting to X64
LINK : error LNK2001: unresolved external symbol mainCRTStartup
a.exe : fatal error LNK1120: 1 unresolved externals
```

I have to admit that the interplay between MSVC, clang & flang here is something that still confuses me.
* clang has a msvc-compatible mode (and then uses MSVC's linker?)
* flang is built on the clang-driver (so presumably could do the same?)
* we're already mixing clang and MSVC here:

https://github.com/conda-forge/scipy-feedstock/blob/3a08a185e651cecdc19023d04c22677b5bdc5d26/recipe/build-output.bat#L57-L59

AFAICT that means we're doing compilation with clang, but linkage with MSVC's linker? Not sure how I'd be able to tell meson to check that, much less in a syntax that flang understands.

Sidenote: If it helps, I have no issues to switch the linker over to lld, that should at least be the same stack as clang/flang.

h-vetinari:
@eli-schwartz, could you maybe share your thoughts how it'd be best to proceed here from your POV? :)

h-vetinari:
OK, using clang as the main compiler on windows allows us to get past the linker check and start compiling. However, we hit a problem pretty immediately:
```
[24/1628] Linking target scipy/_lib/_ccallback_c.cp39-win_amd64.pyd
FAILED: scipy/_lib/_ccallback_c.cp39-win_amd64.pyd 
"clang.exe"  -Wl,/OUT:scipy/_lib/_ccallback_c.cp39-win_amd64.pyd scipy/_lib/_ccallback_c.cp39-win_amd64.pyd.p/meson-generated__ccallback_c.c.obj "-Wl,/nologo" "-Wl,/OPT:REF" "-Wl,/DLL" "-Wl,/IMPLIB:scipy\_lib\_ccallback_c.cp39-win_amd64.lib" "-nostdlib" "-Xclang" "--dependent-lib=msvcrt" "-fuse-ld=lld" "-Wl,-defaultlib:%BUILD_PREFIX%/lib/clang/17.0.0/lib/windows/clang_rt.builtins-x86_64.lib" "-march=nocona" "-mtune=haswell" "-ftree-vectorize" "-fstack-protector-strong" "-O2" "-ffunction-sections" "-pipe" "-D_CRT_SECURE_NO_WARNINGS" "-D_MT" "-D_DLL" "-nostdlib" "-Xclang" "--dependent-lib=msvcrt" "-fuse-ld=lld" "-fno-aligned-allocation" "-Wl,--version-script=C:/bld/scipy-split_1694757427011/work/scipy/_build_utils/link-version-pyinit.map" "%PREFIX%\libs\python39.lib" "-lkernel32" "-luser32" "-lgdi32" "-lwinspool" "-lshell32" "-lole32" "-loleaut32" "-luuid" "-lcomdlg32" "-ladvapi32"
lld-link: warning: ignoring unknown argument '--version-script=C:/bld/scipy-split_1694757427011/work/scipy/_build_utils/link-version-pyinit.map'

lld-link: error: undefined symbol: __declspec(dllimport) strdup
```

strdup [should](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/strdup-wcsdup?view=msvc-170) be in the `ucrt`, and it's present in the host&build env, so I can only imagine that clang is not looking in the right place? Looking at the [content](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fwin-64%2Fucrt-10.0.22621.0-h57928b3_0.tar.bz2), it only contains DLLs and no import libraries - I'm guessing it's possible that those come would with the compiler usually? But the azure image should still contain those... 🤔 

Looking at the last successful build with MSVC for windows for that lib, we see that it's linking to:
```
   INFO (scipy,Lib/site-packages/scipy/_lib/_ccallback_c.cp39-win_amd64.pyd): Needed DSO python39.dll found in conda-forge::python-3.9.18-h4de0772_0_cpython
   INFO (scipy,Lib/site-packages/scipy/_lib/_ccallback_c.cp39-win_amd64.pyd): Needed DSO C:/Windows/System32/KERNEL32.dll found in $SYSROOT
   INFO (scipy,Lib/site-packages/scipy/_lib/_ccallback_c.cp39-win_amd64.pyd): Needed DSO C:/Windows/System32/VCRUNTIME140.dll found in $SYSROOT
   INFO (scipy,Lib/site-packages/scipy/_lib/_ccallback_c.cp39-win_amd64.pyd): Needed DSO C:/Windows/System32/downlevel/api-ms-win-crt-heap-l1-1-0.dll found in $SYSROOT
   INFO (scipy,Lib/site-packages/scipy/_lib/_ccallback_c.cp39-win_amd64.pyd): Needed DSO C:/Windows/System32/downlevel/api-ms-win-crt-runtime-l1-1-0.dll found in $SYSROOT
   INFO (scipy,Lib/site-packages/scipy/_lib/_ccallback_c.cp39-win_amd64.pyd): Needed DSO C:/Windows/System32/downlevel/api-ms-win-crt-string-l1-1-0.dll found in $SYSROOT
   INFO (scipy,Lib/site-packages/scipy/_lib/_ccallback_c.cp39-win_amd64.pyd): Needed DSO C:/Windows/System32/downlevel/api-ms-win-crt-math-l1-1-0.dll found in $SYSROOT
```
So it looks like something about the "sysroot" isn't working for our clang-on-win... @conda-forge/clang-win-activation

h-vetinari:
Beyond the setup questions about `clang_win-64`, I wanted to try building scipy with clang+flang just to see how far flang would get at all. But I realized that we don't have `clang_linux-64` (presumably it's too risky that feedstocks start compiling against it, with a different ABI & stdlib than all our other linux builds).

So I guess I'll retry building flang for osx-64, there at least we'd be able to test the clang+flang combo. And beyond that, I can see what happens with GCC+flang on linux here...

h-vetinari:
Not sure what's happening, but I doubt the GCC+flang combination is well-supported, or even supported at all:
```
[10/1628] Compiling C object scipy/_lib/_test_ccallback.cpython-310-x86_64-linux-gnu.so.p/src__test_ccallback.c.o
x86_64-conda-linux-gnu-cc: warning: $PREFIX/include: linker input file unused because linking not done
x86_64-conda-linux-gnu-cc: warning: $PREFIX/include: linker input file unused because linking not done
x86_64-conda-linux-gnu-cc: warning: $PREFIX/include: linker input file unused because linking not done
x86_64-conda-linux-gnu-cc: warning: $PREFIX/include: linker input file unused because linking not done
[11/1628] Linking target scipy/_lib/_test_ccallback.cpython-310-x86_64-linux-gnu.so
FAILED: scipy/_lib/_test_ccallback.cpython-310-x86_64-linux-gnu.so 
$BUILD_PREFIX/bin/x86_64-conda-linux-gnu-cc  -o scipy/_lib/_test_ccallback.cpython-310-x86_64-linux-gnu.so scipy/_lib/_test_ccallback.cpython-310-x86_64-linux-gnu.so.p/src__test_ccallback.c.o -L$PREFIX/lib -Wl,--as-needed -Wl,--allow-shlib-undefined -Wl,-O1 -shared -fPIC -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,--allow-shlib-undefined -Wl,-rpath,$PREFIX/lib -Wl,-rpath-link,$PREFIX/lib -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,--allow-shlib-undefined -Wl,-rpath,$PREFIX/lib -Wl,-rpath-link,$PREFIX/lib -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe $PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/scipy-split-1.11.2 -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections $PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/scipy-split-1.11.2 -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix -O2 $PREFIX/include -DNDEBUG -D_FORTIFY_SOURCE=2 -O2 $PREFIX/include
$BUILD_PREFIX/bin/../lib/gcc/x86_64-conda-linux-gnu/12.3.0/../../../../x86_64-conda-linux-gnu/bin/ld: error: $PREFIX/include: read: Is a directory
collect2: error: ld returned 1 exit status
```

eli-schwartz:
That looks suspiciously like something was supposed to normalize paths but is actually chopping off the `-I` part, and now the compiler thinks that an include headers path is actually an object file inpu.

h-vetinari:
So with https://github.com/conda-forge/flang-feedstock/pull/29, the intention is to test whether flang can actually compile scipy's fortran code (while we figure out the windows side of things; this would at least allow raising issues upstream in case there are any major issues).

For now, we cannot yet use flang on osx until https://github.com/conda-forge/pythran-feedstock/pull/80 is merged.

h-vetinari:
Flang-on-osx is also running into some issues with `$PREFIX/include` it seems:
```
[15/1628] Linking target scipy/_lib/_fpumode.cpython-311-darwin.so
FAILED: scipy/_lib/_fpumode.cpython-311-darwin.so 
x86_64-apple-darwin13.4.0-clang  -o scipy/_lib/_fpumode.cpython-311-darwin.so scipy/_lib/_fpumode.cpython-311-darwin.so.p/_fpumode.c.o -L$PREFIX/lib -Wl,-dead_strip_dylibs -Wl,-headerpad_max_install_names -Wl,-undefined,dynamic_lookup -bundle -Wl,-undefined,dynamic_lookup -Wl,-pie -Wl,-headerpad_max_install_names -Wl,-dead_strip_dylibs -Wl,-rpath,$PREFIX/lib -march=core2 -mtune=haswell -mssse3 -ftree-vectorize -fPIC -fPIE -fstack-protector-strong -O2 -pipe $PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/scipy-split-1.11.2 -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix -D_FORTIFY_SOURCE=2 $PREFIX/include -mmacosx-version-min=10.9
ld: warning: -pie being ignored. It is only used when linking a main executable
ld: can't map file, errno=22 file '$PREFIX/include' for architecture x86_64
x86_64-apple-darwin13.4: error: linker command failed with exit code 1 (use -v to see invocation)
```

h-vetinari:
Ah, one more thing to override in meson:
```
lld-link: error: could not open 'flang.lib': no such file or directory
lld-link: error: could not open 'pgmath.lib': no such file or directory
```

h-vetinari:
Alriiiiiiiight, now we're cookin' 😎 

Fortan compilation is working so far, failure is actually happening in the C++ bits, presumably due to clang-vs-STL or something:
```
[837/1628] Compiling C++ object scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/meson-generated_binom_ufunc.cpp.obj
FAILED: scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/meson-generated_binom_ufunc.cpp.obj 
"clang++.exe" "-Iscipy\stats\_boost\binom_ufunc.cp311-win_amd64.pyd.p" "-Iscipy\stats\_boost" "-I..\scipy\stats\_boost" "-I..\scipy\stats\_boost\include" "-I..\scipy\_lib\boost_math\include" "-I..\..\_h_env\Lib\site-packages\numpy\core\include" "-I%PREFIX%\Include" "-fvisibility=hidden" "-fvisibility-inlines-hidden" "-fcolor-diagnostics" "-DNDEBUG" "-D_FILE_OFFSET_BITS=64" "-Wall" "-Winvalid-pch" "-std=c++14" "-O3" "-fvisibility-inlines-hidden" "-std=c++17" "-fmessage-length=0" "-march=nocona" "-mtune=haswell" "-ftree-vectorize" "-fstack-protector-strong" "-O2" "-ffunction-sections" "-pipe" "-D_CRT_SECURE_NO_WARNINGS" "-D_MT" "-D_DLL" "-nostdlib" "-Xclang" "--dependent-lib=msvcrt" "-fuse-ld=lld" "-fno-aligned-allocation" "-DMS_WIN64=" "-DBOOST_MATH_STANDALONE=1" "-DBOOST_MATH_DOMAIN_ERROR_POLICY=ignore_error" "-DBOOST_MATH_EVALUATION_ERROR_POLICY=user_error" "-DBOOST_MATH_OVERFLOW_ERROR_POLICY=user_error" "-DBOOST_MATH_PROMOTE_DOUBLE_POLICY=false" "-DNPY_NO_DEPRECATED_API=NPY_1_9_API_VERSION" "-Wno-cpp" "-D_USE_MATH_DEFINES" -MD -MQ scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/meson-generated_binom_ufunc.cpp.obj -MF "scipy\stats\_boost\binom_ufunc.cp311-win_amd64.pyd.p\meson-generated_binom_ufunc.cpp.obj.d" -o scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/meson-generated_binom_ufunc.cpp.obj "-c" scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/binom_ufunc.cpp
In file included from scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/binom_ufunc.cpp:1176:
C:/bld/scipy-split_1695344021630/work/scipy/stats/_boost/include/func_defs.hpp:18:1: error: target of using declaration conflicts with declaration already in scope
   18 | BOOST_MATH_DECLARE_SPECIAL_FUNCTIONS(user_error_policy)
      | ^
..\scipy\_lib\boost_math\include\boost/math/special_functions/math_fwd.hpp:1577:23: note: expanded from macro 'BOOST_MATH_DECLARE_SPECIAL_FUNCTIONS'
 1577 |    using boost::math::isfinite;\
      |                       ^
..\scipy\_lib\boost_math\include\boost/math/special_functions/fpclassify.hpp:355:14: note: target of using declaration
  355 | inline bool (isfinite)(T x)
      |              ^
C:\Program Files (x86)\Windows Kits\10\include\10.0.22621.0\ucrt\corecrt_math.h:405:32: note: conflicting declaration
  405 |     _Check_return_ inline bool isfinite(_In_ _Ty _X) throw()
      |                                ^
```

h-vetinari:
Not sure why that error triggers only on clang and not MSVC... In any case, I cannot investigate further until next week...

rgommers:
That looks like an issue in the boost-math version included in scipy:

```
[837/1628] Compiling C++ object scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/meson-generated_binom_ufunc.cpp.obj
FAILED: scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/meson-generated_binom_ufunc.cpp.obj 
"clang++.exe" "-Iscipy\stats\_boost\binom_ufunc.cp311-win_amd64.pyd.p" "-Iscipy\stats\_boost" "-I..\scipy\stats\_boost" "-I..\scipy\stats\_boost\include" "-I..\scipy\_lib\boost_math\include" "-I..\..\_h_env\Lib\site-packages\numpy\core\include" "-I%PREFIX%\Include" "-fvisibility=hidden" "-fvisibility-inlines-hidden" "-fcolor-diagnostics" "-DNDEBUG" "-D_FILE_OFFSET_BITS=64" "-Wall" "-Winvalid-pch" "-std=c++14" "-O3" "-fvisibility-inlines-hidden" "-std=c++17" "-fmessage-length=0" "-march=nocona" "-mtune=haswell" "-ftree-vectorize" "-fstack-protector-strong" "-O2" "-ffunction-sections" "-pipe" "-D_CRT_SECURE_NO_WARNINGS" "-D_MT" "-D_DLL" "-nostdlib" "-Xclang" "--dependent-lib=msvcrt" "-fuse-ld=lld" "-fno-aligned-allocation" "-DBOOST_MATH_USE_STD_FPCLASSIFY" "-DMS_WIN64=" "-DBOOST_MATH_STANDALONE=1" "-DBOOST_MATH_DOMAIN_ERROR_POLICY=ignore_error" "-DBOOST_MATH_EVALUATION_ERROR_POLICY=user_error" "-DBOOST_MATH_OVERFLOW_ERROR_POLICY=user_error" "-DBOOST_MATH_PROMOTE_DOUBLE_POLICY=false" "-DNPY_NO_DEPRECATED_API=NPY_1_9_API_VERSION" "-Wno-cpp" "-D_USE_MATH_DEFINES" -MD -MQ scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/meson-generated_binom_ufunc.cpp.obj -MF "scipy\stats\_boost\binom_ufunc.cp311-win_amd64.pyd.p\meson-generated_binom_ufunc.cpp.obj.d" -o scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/meson-generated_binom_ufunc.cpp.obj "-c" scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/binom_ufunc.cpp
In file included from scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/binom_ufunc.cpp:1176:
C:/bld/scipy-split_1695349531254/work/scipy/stats/_boost/include/func_defs.hpp:18:1: error: target of using declaration conflicts with declaration already in scope
   18 | BOOST_MATH_DECLARE_SPECIAL_FUNCTIONS(user_error_policy)
      | ^
..\scipy\_lib\boost_math\include\boost/math/special_functions/math_fwd.hpp:1579:23: note: expanded from macro 'BOOST_MATH_DECLARE_SPECIAL_FUNCTIONS'
 1579 |    using boost::math::isnan;\
      |                       ^
..\scipy\_lib\boost_math\include\boost/math/special_functions/fpclassify.hpp:605:14: note: target of using declaration
  605 | inline bool (isnan)(T x)
      |              ^
C:\Program Files (x86)\Windows Kits\10\include\10.0.22621.0\ucrt\corecrt_math.h:417:32: note: conflicting declaration
  417 |     _Check_return_ inline bool isnan(_In_ _Ty _X) throw()
      |                                ^
In file included from scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/binom_ufunc.cpp:1176:
C:/bld/scipy-split_1695349531254/work/scipy/stats/_boost/include/func_defs.hpp:18:1: error: target of using declaration conflicts with declaration already in scope
   18 | BOOST_MATH_DECLARE_SPECIAL_FUNCTIONS(user_error_policy)
      | ^
..\scipy\_lib\boost_math\include\boost/math/special_functions/math_fwd.hpp:1580:23: note: expanded from macro 'BOOST_MATH_DECLARE_SPECIAL_FUNCTIONS'
 1580 |    using boost::math::isnormal;\
      |                       ^
..\scipy\_lib\boost_math\include\boost/math/special_functions/fpclassify.hpp:426:14: note: target of using declaration
  426 | inline bool (isnormal)(T x)
      |              ^
C:\Program Files (x86)\Windows Kits\10\include\10.0.22621.0\ucrt\corecrt_math.h:423:32: note: conflicting declaration

  423 |     _Check_return_ inline bool isnormal(_In_ _Ty _X) throw()

      |                                ^

4 errors generated.
```

I don't see many changes in the problematic files: https://github.com/boostorg/math/blob/09d82da6e9884da86495279b331ff038dd1330b8/include/boost/math/special_functions/fpclassify.hpp#L605. There are some usages of `_MSC_VER`, but Clang-cl defines that so that doesn't explain why things work with MSVC but not clang-cl. 

That code hasn't changes a lot, the only potentially relevant commit I see is https://github.com/boostorg/math/commit/dbb1ae7667feb82ace8b25a6cf30e5fd830094b4. There are also a lot of Boost-specific defines (e.g., `-DBOOST_USE_WINDOWS_H`) that may be used, but it's tricky to figure out if one is relevant or should be set by default in scipy's `meson.build` files.

@mborland maybe the above looks familiar to you?

mborland:
> That looks like an issue in the boost-math version included in scipy:
> 
> ```
> [837/1628] Compiling C++ object scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/meson-generated_binom_ufunc.cpp.obj
> FAILED: scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/meson-generated_binom_ufunc.cpp.obj 
> "clang++.exe" "-Iscipy\stats\_boost\binom_ufunc.cp311-win_amd64.pyd.p" "-Iscipy\stats\_boost" "-I..\scipy\stats\_boost" "-I..\scipy\stats\_boost\include" "-I..\scipy\_lib\boost_math\include" "-I..\..\_h_env\Lib\site-packages\numpy\core\include" "-I%PREFIX%\Include" "-fvisibility=hidden" "-fvisibility-inlines-hidden" "-fcolor-diagnostics" "-DNDEBUG" "-D_FILE_OFFSET_BITS=64" "-Wall" "-Winvalid-pch" "-std=c++14" "-O3" "-fvisibility-inlines-hidden" "-std=c++17" "-fmessage-length=0" "-march=nocona" "-mtune=haswell" "-ftree-vectorize" "-fstack-protector-strong" "-O2" "-ffunction-sections" "-pipe" "-D_CRT_SECURE_NO_WARNINGS" "-D_MT" "-D_DLL" "-nostdlib" "-Xclang" "--dependent-lib=msvcrt" "-fuse-ld=lld" "-fno-aligned-allocation" "-DBOOST_MATH_USE_STD_FPCLASSIFY" "-DMS_WIN64=" "-DBOOST_MATH_STANDALONE=1" "-DBOOST_MATH_DOMAIN_ERROR_POLICY=ignore_error" "-DBOOST_MATH_EVALUATION_ERROR_POLICY=user_error" "-DBOOST_MATH_OVERFLOW_ERROR_POLICY=user_error" "-DBOOST_MATH_PROMOTE_DOUBLE_POLICY=false" "-DNPY_NO_DEPRECATED_API=NPY_1_9_API_VERSION" "-Wno-cpp" "-D_USE_MATH_DEFINES" -MD -MQ scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/meson-generated_binom_ufunc.cpp.obj -MF "scipy\stats\_boost\binom_ufunc.cp311-win_amd64.pyd.p\meson-generated_binom_ufunc.cpp.obj.d" -o scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/meson-generated_binom_ufunc.cpp.obj "-c" scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/binom_ufunc.cpp
> In file included from scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/binom_ufunc.cpp:1176:
> C:/bld/scipy-split_1695349531254/work/scipy/stats/_boost/include/func_defs.hpp:18:1: error: target of using declaration conflicts with declaration already in scope
>    18 | BOOST_MATH_DECLARE_SPECIAL_FUNCTIONS(user_error_policy)
>       | ^
> ..\scipy\_lib\boost_math\include\boost/math/special_functions/math_fwd.hpp:1579:23: note: expanded from macro 'BOOST_MATH_DECLARE_SPECIAL_FUNCTIONS'
>  1579 |    using boost::math::isnan;\
>       |                       ^
> ..\scipy\_lib\boost_math\include\boost/math/special_functions/fpclassify.hpp:605:14: note: target of using declaration
>   605 | inline bool (isnan)(T x)
>       |              ^
> C:\Program Files (x86)\Windows Kits\10\include\10.0.22621.0\ucrt\corecrt_math.h:417:32: note: conflicting declaration
>   417 |     _Check_return_ inline bool isnan(_In_ _Ty _X) throw()
>       |                                ^
> In file included from scipy/stats/_boost/binom_ufunc.cp311-win_amd64.pyd.p/binom_ufunc.cpp:1176:
> C:/bld/scipy-split_1695349531254/work/scipy/stats/_boost/include/func_defs.hpp:18:1: error: target of using declaration conflicts with declaration already in scope
>    18 | BOOST_MATH_DECLARE_SPECIAL_FUNCTIONS(user_error_policy)
>       | ^
> ..\scipy\_lib\boost_math\include\boost/math/special_functions/math_fwd.hpp:1580:23: note: expanded from macro 'BOOST_MATH_DECLARE_SPECIAL_FUNCTIONS'
>  1580 |    using boost::math::isnormal;\
>       |                       ^
> ..\scipy\_lib\boost_math\include\boost/math/special_functions/fpclassify.hpp:426:14: note: target of using declaration
>   426 | inline bool (isnormal)(T x)
>       |              ^
> C:\Program Files (x86)\Windows Kits\10\include\10.0.22621.0\ucrt\corecrt_math.h:423:32: note: conflicting declaration
> 
>   423 |     _Check_return_ inline bool isnormal(_In_ _Ty _X) throw()
> 
>       |                                ^
> 
> 4 errors generated.
> ```
> 
> I don't see many changes in the problematic files: https://github.com/boostorg/math/blob/09d82da6e9884da86495279b331ff038dd1330b8/include/boost/math/special_functions/fpclassify.hpp#L605. There are some usages of `_MSC_VER`, but Clang-cl defines that so that doesn't explain why things work with MSVC but not clang-cl.
> 
> That code hasn't changes a lot, the only potentially relevant commit I see is [boostorg/math@dbb1ae7](https://github.com/boostorg/math/commit/dbb1ae7667feb82ace8b25a6cf30e5fd830094b4). There are also a lot of Boost-specific defines (e.g., `-DBOOST_USE_WINDOWS_H`) that may be used, but it's tricky to figure out if one is relevant or should be set by default in scipy's `meson.build` files.
> 
> @mborland maybe the above looks familiar to you?

These kinds of ADL failures and relatively common with MSVC so I guess I am not surprised it would extend to Clang-cl. I've added clang-cl to our CI now (https://github.com/boostorg/math/pull/1030) to see if we root out the failure.

cc: @jzmaddock

jzmaddock:
Where can I see the source for func_defs.hpp?  Is BOOST_MATH_DECLARE_SPECIAL_FUNCTIONS being used at global scope (a really really bad idea)?

rgommers:
> Where can I see the source for func_defs.hpp?

https://github.com/scipy/scipy/blob/main/scipy/stats/_boost/include/func_defs.hpp

> Is BOOST_MATH_DECLARE_SPECIAL_FUNCTIONS being used at global scope (a really really bad idea)?

Looks like it. It's in global scope in `func_defs.hpp`, and that's included in global scope in each generated C++ source file for a Python extension module. E.g., from `beta_ufunc.cpp` (generated by Cython via a `.pxd` include which is itself generated by [stats/_boost/include/gen_func_defs_pxd.py](https://github.com/scipy/scipy/blob/main/scipy/stats/_boost/include/gen_func_defs_pxd.py), so I can't link to the source):

<details>

```
/* Early includes */
#include <string.h>
#include <stdio.h>
#include "numpy/arrayobject.h"
#include "numpy/ndarrayobject.h"
#include "numpy/ndarraytypes.h"
#include "numpy/arrayscalars.h"
#include "numpy/ufuncobject.h"

    /* NumPy API declarations from "numpy/__init__.pxd" */
    
#include "ios"
#include "new"
#include "stdexcept"
#include "typeinfo"
#include "Templated_PyUFunc.hpp"
#include "/abspath/to/scipy/stats/_boost/include/func_defs.hpp"
#include "boost/math/distributions/beta.hpp"
#ifdef _OPENMP
#include <omp.h>
#endif /* _OPENMP */
```

</details>

jzmaddock:
>Looks like it. It's in global scope in func_defs.hpp, and that's included in global scope in each generated C++ source file for a Python extension module. 

That's the cause of the issue then: the std lib and Boost end up defining the same function name in global scope.

rgommers:
Thank you @jzmaddock, that was very helpful. I think https://github.com/scipy/scipy/pull/19286 should fix this. 

There may be more to do to clean up `func_defs.hpp`; after reading the docs at https://www.boost.org/doc/libs/1_83_0/libs/math/doc/html/math_toolkit/pol_overview.html it looks to me like that whole file may be due for some restructuring - there's no namespace usage, and the way of applying patches to distribution methods is hard to understand.

h-vetinari:
Thanks everyone for the quick responses, inputs and fixes! 🙏 

Backporting https://github.com/scipy/scipy/pull/19286, compilation now progresses past the boost submodule, so that looks great!

Next issue seems to be some path mangling problem (forward vs. backward slash maybe), I don't know where yet:
```
[1022/1628] Linking target scipy/stats/_unuran/unuran_wrapper.cp311-win_amd64.pyd
FAILED: scipy/stats/_unuran/unuran_wrapper.cp311-win_amd64.pyd 
"clang.exe" @scipy/stats/_unuran/unuran_wrapper.cp311-win_amd64.pyd.rsp
clang: error: no such file or directory: 'C:bldscipy-split_1695543774951_h_envlibspython311.lib'
```

rgommers:
Perhaps a hiccup with Meson's Clang on Windows config, because it does work with MSVC and Mingw. However, the `unuran_wrapper` config in SciPy is quite odd, so this should be easy to fix in SciPy. I think https://github.com/scipy/scipy/pull/19289 should do it.

h-vetinari:
I'm still not 100% sure how we'll want to go about `clang.exe` vs. `clang-cl.exe`. Presumably we need the latter for ABI-compat with the rest of our windows-packages. Switching to that now fails a bunch earlier than vanilla clang:
```
scipy/_lib/messagestream.cp311-win_amd64.pyd.p/messagestream.c(3608,34): error: incompatible pointer to integer conversion passing 'const char *' to parameter of type 'int' [-Wint-conversion]
 3608 |   __pyx_v_self->handle = _fdopen(__pyx_t_11, ((char const *)"wb+"));
      |                                  ^~~~~~~~~~
C:\Program Files (x86)\Windows Kits\10\include\10.0.22621.0\ucrt\stdio.h(161,28): note: passing argument to parameter '_FileHandle' here
  161 |         _In_   int         _FileHandle,
      |                            ^
```
This seems to be a problem involving Cython-generated C-code, clang's relatively tight warnings, plus the way MSFT wrote their C stdlib...

But it's also pretty clearly related to the patch that @xhochy came up with for compatibility with vanilla clang (+ ucrt), so for now I'm trying without that.

h-vetinari:
> @rgommers: [...] I think [scipy/scipy#19289](https://github.com/scipy/scipy/pull/19289) should do it.

Awesome! I think I'll open another PR for continuing on the vanilla clang approach (which would be great to debug as well, if only to see that we can compile a functional package).

rgommers:
> I'm still not 100% sure how we'll want to go about `clang.exe` vs. `clang-cl.exe`. Presumably we need the latter for ABI-compat with the rest of our windows-packages.

I think so too, we need `clang-cl` - SciPy has some C/C++ code that uses `long double`, and there be some other hiccups with plain `clang` as well. Happy to work through this and make both work of course.

rgommers:
Compiler selection and activation seem to be a bit misaligned, this should be using MSVC flag format:
```
clang-cl: warning: unknown argument ignored in clang-cl: '-fvisibility-inlines-hidden' [-Wunknown-argument]
clang-cl: warning: unknown argument ignored in clang-cl: '-std=c++17' [-Wunknown-argument]
clang-cl: warning: unknown argument ignored in clang-cl: '-fmessage-length=0' [-Wunknown-argument]
clang-cl: warning: unknown argument ignored in clang-cl: '-mtune=haswell' [-Wunknown-argument]
clang-cl: warning: unknown argument ignored in clang-cl: '-ftree-vectorize' [-Wunknown-argument]
clang-cl: warning: unknown argument ignored in clang-cl: '-fstack-protector-strong' [-Wunknown-argument]
clang-cl: warning: unknown argument ignored in clang-cl: '-ffunction-sections' [-Wunknown-argument]
clang-cl: warning: unknown argument ignored in clang-cl: '-pipe' [-Wunknown-argument]
clang-cl: warning: unknown argument ignored in clang-cl: '-nostdlib' [-Wunknown-argument]
clang-cl: warning: unknown argument ignored in clang-cl: '-fno-aligned-allocation' [-Wunknown-argument]
```

h-vetinari:
Alright, next milestone: the first windows+meson build that passes compilation 🥳 

Now failing upon installation:
```
  File "C:\bld\scipy-split_1695552615240\_h_env\Lib\site-packages\mesonpy\_wheelfile.py", line 94, in write
    with open(filename, 'rb') as f:
         ^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\bld\\scipy-split_1695552615240\\work\\builddir\\scipy\\special\\_specfun.cp311-win_amd64.dll.a'

ERROR Backend subprocess exited when trying to invoke build_wheel
```

Looking further up the logs, this is because `lld` does not deal with `--out-implib` (or, in fact, many of the other CLI arguments)
```
[255/1628] Linking target scipy/special/_specfun.cp311-win_amd64.pyd
lld-link: warning: ignoring unknown argument '--allow-shlib-undefined'
lld-link: warning: ignoring unknown argument '-O1'
lld-link: warning: ignoring unknown argument '--start-group'
lld-link: warning: ignoring unknown argument '--out-implib=scipy\special\_specfun.cp311-win_amd64.dll.a'
lld-link: warning: ignoring unknown argument '--end-group'
[256/1628] Compiling C++ object scipy/special/_ufuncs_cxx.cp311-win_amd64.pyd.p/meson-generated__ufuncs_cxx.cpp.obj
```

h-vetinari:
> Compiler selection and activation seem to be a bit misaligned

Yeah, this should probably just use the vc-activation. And we might not want to use lld after all? 🤷 

I'll push that change, but cannot babysit it further tonight...

h-vetinari:
> Compiler selection and activation seem to be a bit misaligned

The problem is that, even with `CC=clang-cl.exe`, meson will stumble over finding the linker associated with flang, if we use the windows activation:
```
..\meson.build:82:0: ERROR: Unable to detect linker for compiler `flang-new -Wl,--version`
stdout: LINK : warning LNK4044: unrecognized option '/-version'; ignored
```

I see that some fortran compilers in [meson](https://github.com/mesonbuild/meson/blob/1.2.1/mesonbuild/compilers/fortran.py) have
```
LINKER_PREFIX = '-Wl,'
```
but AFAICT that's not the problem (i.e. `--version` already gets passed to `LINK`, which then stumbles over that), and I don't see where I'd change `--version` for the linker detection. From a quick look at the [meson](https://mesonbuild.com/Using-with-Visual-Studio.html#using-clangcl-with-visual-studio) [docs](https://mesonbuild.com/Reference-tables.html#compiler-ids), we might need a flang-variant that uses msvc argument syntax?

h-vetinari:
So I managed to bypass the linker check by setting `FC_LD=lld-link.exe`, now we're failing with
```
[257/1628] Linking target scipy/special/_specfun.cp311-win_amd64.pyd
FAILED: scipy/special/_specfun.cp311-win_amd64.pyd 
"flang-new"  -o scipy/special/_specfun.cp311-win_amd64.pyd scipy/special/_specfun.cp311-win_amd64.pyd.p/meson-generated_..__specfunmodule.c.obj "-LC:/bld/scipy-split_1695593746883/_build_env/Library/lib/clang/17" "-Wl,--allow-shlib-undefined" "-fuse-ld=lld-link.exe" "-Wl,-O1" "-shared" "-Wl,--start-group" "-Wl,--out-implib=scipy\special\_specfun.cp311-win_amd64.dll.a" "-Wl,-Lucrt" "-DAVOID_NATIVE_UINT128_T=1" "scipy\special\libspecfun.a" "scipy\lib_fortranobject.a" "%PREFIX%\libs\python311.lib" "-lFortranRuntime" "-lFortranDecimal" "-Wl,--end-group"
lld-link: warning: ignoring unknown argument '--allow-shlib-undefined'
lld-link: warning: ignoring unknown argument '-O1'
lld-link: warning: ignoring unknown argument '--start-group'
lld-link: warning: ignoring unknown argument '--out-implib=scipy\special\_specfun.cp311-win_amd64.dll.a'
lld-link: warning: ignoring unknown argument '-Lucrt'

lld-link: warning: ignoring unknown argument '--end-group'

lld-link: error: undefined symbol: __divdc3
```
presumably because `-Lucrt` isn't being picked up.

So it still seems we need to tell meson that the linker needs MSVC-style arguments. Any pointers how we could achieve that @eli-schwartz?

isuruf:
`-Lucrt` is about giving a folder. You have to use the options as done in https://github.com/conda-forge/clang-win-activation-feedstock/blob/main/recipe/activate-clang_win-64.sh#L111C15-L111C150

isuruf:
Particularly `-Xclang`

h-vetinari:
> `-Lucrt` is about giving a folder.

Yeah, that was my understanding too, but I didn't touch it yet because it wasn't my commit (https://github.com/conda-forge/scipy-feedstock/pull/246/commits/cd3dde0f78b209d7fdc462c94b62af2da5758938).

The problem of meson passing gnu-style args to lld-link is an issue though, particularly with `--out-implib` not being picked up, because meson then expects those files to be present upon installation.

It might work to use `ld.lld` for flang though? That should be compatible as well with the activation arguments you gave, and should match the current expectations of meson at least.

In any case, thanks a lot for the inputs!

I'll admit that it's not easy for me to figure out the right choice of components and how they all interplay; from clang(-cl) to lld(-link) to flang to meson to ucrt/compiler-rt, but so far:
* ❌ clang & ... (wrong ABI)
* ❌ clang-cl & lld-link (meson uses wrong CLI style)
* ❓ clang-cl & lld-link & meson-patches (patch how...?)
* ❓ clang-cl & ld.lld (might work?)

h-vetinari:
Interestingly, meson falls on its face with `CC_LD=ld.lld` on windows...

h-vetinari:
Well... compilation passes again, but the build still fails due to the ignored `--out-implib` in
```
lld-link: warning: ignoring unknown argument '--out-implib=scipy\special\_specfun.cp311-win_amd64.dll.a'
```
Obviously the respective file(s) that meson expects to copy are then missing upon installation.

This puts us in a bit of a bind, because `lld-link` (for CC/CXX) ignores the relevant CLI arguments, but using `ld.lld` (which would understand the arguments) + `clang-cl` on windows [crashes](https://github.com/conda-forge/scipy-feedstock/pull/246#issuecomment-1732713176) meson.

h-vetinari:
@eli-schwartz @dcbaker @jpakkane @xclaesse
I'm planning on opening some meson issues; perhaps you have an opinion which ones are more worthwhile / relevant than others:
* `lld-link` [using](https://github.com/conda-forge/scipy-feedstock/pull/246#issuecomment-1732687428) GNU syntax for the CLI arguments (which leads to failures because `--out-implib` gets ignored, yet installation expects the respective files).
  * working around this by setting `CC=clang-cl.exe` & `CC_LD=ld.lld` [crashes](https://github.com/conda-forge/scipy-feedstock/pull/246#issuecomment-1732713176) meson
* meson [assuming](https://github.com/conda-forge/scipy-feedstock/pull/253#issuecomment-1732683380) `can_linker_accept_rsp() == True` on windows, even though lld does not support it (apparently?)
  * alternative: rsp handling wrongly [mangles](https://github.com/conda-forge/scipy-feedstock/pull/253#issuecomment-1732545515) paths somewhere
  * `MESON_RSP_THRESHOLD` not [working](https://github.com/conda-forge/scipy-feedstock/pull/253#issuecomment-1732578945) as expected (at least on windows with `clang.exe`)
* general support for `llvm-flang`:
  * probably a longer discussion (because some things are in flight in development and subject to change); minimal changes currently (for flang 17) look like [this](https://github.com/conda-forge/meson-feedstock/blob/dev/recipe/patches/0002-minimal-adaptations-for-llvm-flang.patch)
  * linker detection doesn't work correctly, currently needs `FC_LD`

h-vetinari:
> linker detection doesn't work correctly, currently needs `FC_LD`

I think this might be the root of the issues here. Even when I try to work around the non-functional linker detection for flang on windows by setting `FC_LD=lld-link`, I still end up with
```
Fortran compiler for the host machine: flang-new (flang 17.0.0 "flang-new version 17.0.0")
Fortran linker for the host machine: flang-new ld.lld 17.0.0
                                               ^^^^^^
                                               wrong!
```
This probably explains why the syntax for the linker is not the right one, and all the problems that follow from that.

I haven't managed to fully dig through the list of class inheritance to understand where the linker detection happens for flang, but if I understand the code for [`guess_win_linker`](https://github.com/mesonbuild/meson/blob/1.2.1/mesonbuild/linkers/detect.py#L47) correctly, it specifically looks for a string `LLD` in the compiler version output, but neither clang-cl nor flang-new have that:
```
(%BUILD_PREFIX%) %SRC_DIR%>clang-cl.exe --version 
clang version 17.0.0
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: %BUILD_PREFIX%\Library\bin

(%BUILD_PREFIX%) %SRC_DIR%>flang-new.exe --version 
flang-new version 17.0.0
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: %BUILD_PREFIX%\Library\bin
```

Somehow, on clang things work (and lld-link) is picked up, not sure where the respective code is. If I set a full path like  or `FD_LD=%FULL_PATH%\lld-link.exe`, the code falls through (AFAICT) to `ClangCompiler.use_linker_args` and then `GnuLikeCompiler.use_linker_args`, where it fails with:
```
..\meson.build:82:0: ERROR: Unsupported linker, only bfd, gold, and lld are supported, not %PREFIX%\Library\bin\lld-link.exe.
```
I'll try [overriding](https://github.com/conda-forge/meson-feedstock/pull/88) `use_linker_args` in `FlangCompiler`, but that doesn't seem to be where the unix-y flavour of `lld` is being picked up...?

h-vetinari:
> See also the `invoked_directly` flag.

I think I see the issue now. `detect_fortran_compiler` only [does](https://github.com/mesonbuild/meson/blob/32be9bd6de0a4769cdf6d9528496925643739918/mesonbuild/compilers/detect.py#L784) `guess_nix_linker` for flang. That also explains the how the C/C++ compilers pick up lld-link, because of the [try-except](https://github.com/mesonbuild/meson/blob/32be9bd6de0a4769cdf6d9528496925643739918/mesonbuild/compilers/detect.py#L456) with `invoked_directly=False` on windows.

h-vetinari:
Woohoo 🥳 
```
Fortran compiler for the host machine: flang-new (flang 17.0.0 "flang-new version 17.0.0")
Fortran linker for the host machine: flang-new lld-link 17.0.0
```

h-vetinari:
Ah, the joy was shortlived. Now we have the rsp issue also for flang:
```
[353/1628] Linking target scipy/linalg/_interpolative.cp311-win_amd64.pyd
FAILED: scipy/linalg/_interpolative.cp311-win_amd64.pyd 
"flang-new" @scipy/linalg/_interpolative.cp311-win_amd64.pyd.rsp
flang-new: error: no such file or directory: 'scipylib_fortranobject.a'
flang-new: error: no such file or directory: 'C:bldscipy-split_1695940085251_h_envlibspython311.lib'
```

h-vetinari:
Aaaaand [marking](https://github.com/conda-forge/meson-feedstock/pull/90) rsp as unsupported for lld-link causes:
```
..\meson.build:82:0: ERROR: Compiler flang-new cannot compile programs.
```
😑

h-vetinari:
> @isuruf: You have to use the options as done in https://github.com/conda-forge/clang-win-activation-feedstock/blob/2f9e6d3774c9a702bc9b48ccc2ad83d0e7ea64b1/recipe/activate-clang_win-64.sh#L116
>
> Particularly `-nostdlib -Xclang --dependent-lib=msvcrt`

From the meson logs:
```
clang-cl: warning: unknown argument ignored in clang-cl: '-nostdlib' [-Wunknown-argument]
clang-cl: warning: unknown argument ignored in clang-cl: '-fno-aligned-allocation' [-Wunknown-argument]
lld-link: warning: ignoring unknown argument '-Wl,-defaultlib:%PREFIX%\Library/lib/clang/17.0.0/lib/windows/clang_rt.builtins-x86_64.lib'
```
It looks like this affects regular C compilation as well:
```
[14/1628] Linking target scipy/_lib/_fpumode.cp311-win_amd64.pyd
lld-link: warning: ignoring unknown argument '--target=x86_64-pc-windows-msvc'
lld-link: warning: ignoring unknown argument '-nostdlib'
lld-link: warning: ignoring unknown argument '-Xclang'
lld-link: warning: ignoring unknown argument '--dependent-lib=msvcrt'
lld-link: warning: ignoring unknown argument '-fuse-ld=lld'
lld-link: warning: ignoring unknown argument '-Wl,-defaultlib:%PREFIX%\Library/lib/clang/17.0.0/lib/windows/clang_rt.builtins-x86_64.lib'
```

Could it be that we have to adapt the activation for windows to use `clang-cl` (& respective syntax) rather than `clang`?

Similarly, flang ignores `--dependent-lib`
```
flang-new: warning: -Wl,-defaultlib:%PREFIX%\Library/lib/clang/17.0.0/lib/windows/clang_rt.builtins-x86_64.lib: 'linker' input unused
flang-new: warning: argument unused during compilation: '-Xclang --dependent-lib=msvcrt'
```
resp. [bails](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=791496&view=results) when it's done through `-Xflang`:
```
error: unknown argument: '--dependent-lib=msvcrt'
```
What are your thoughts on dealing with this? Can we get away without using `--dependent-lib` for flang (probably means we'd be linking to compiler-rt, but then, we're already encouraging that in `LDFLAGS` anyway...)?

PS. I [raised](https://github.com/llvm/llvm-project/issues/63741) this a while ago upstream

h-vetinari:
> `MESON_RSP_THRESHOLD` not [working](https://github.com/conda-forge/scipy-feedstock/pull/253#issuecomment-1732578945) as expected

OK, it does seem to work with a ridiculously high threshold (used 320'000 instead of 32'000).

That means we passed the first build! 🥳

h-vetinari:
# IT'S GREEN!!! 🤩 🥳 🚀 

giphy

rgommers:
🚀 🎉

PR #253, opened 2023-09-24
WIP: meson on windows (clang.exe)

h-vetinari:
Version of #246 with vanilla clang (not the MSVC-compatible `clang-cl.exe`).


h-vetinari:
@rgommers
https://github.com/scipy/scipy/pull/19289 did not change the error for unuran:
```
[1023/1629] Linking target scipy/stats/_unuran/unuran_wrapper.cp311-win_amd64.pyd
FAILED: scipy/stats/_unuran/unuran_wrapper.cp311-win_amd64.pyd 
"clang.exe" @scipy/stats/_unuran/unuran_wrapper.cp311-win_amd64.pyd.rsp
clang: error: no such file or directory: 'C:bldscipy-split_1695551724191_h_envlibspython311.lib'
```

h-vetinari:
> So probably this can be worked around by setting `MESON_RSP_THRESHOLD` to a value close to 32k?

That sounds like an excellent analysis, but unfortunately it didn't seem to change anything...

It seems that meson has [metadata](https://github.com/mesonbuild/meson/blob/1.2.1/mesonbuild/compilers/compilers.py#L636-L640) for the linker to advertise rsp compatibility, and it seems that just falls through to the [default](https://github.com/mesonbuild/meson/blob/1.2.1/mesonbuild/linkers/linkers.py#L404-L407), which is set to always true on windows.

I'm going to try patching that in https://github.com/conda-forge/meson-feedstock/pull/86

h-vetinari:
> I'm going to try patching that in [conda-forge/meson-feedstock#86](https://github.com/conda-forge/meson-feedstock/pull/86)

That didn't seem to have any effect either 😑

rgommers:
Then I'm out of workaround ideas - I guess it needs a proper fix in Meson. Unfortunately I don't have a Windows machine at hand, and this isn't a debug-via-CI kinda problem.

PR #252, opened 2023-09-19
Win meson clang

xhochy:
Checklist
* [ ] Used a [personal fork of the feedstock to propose changes](https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests)
* [ ] Bumped the build number (if the version is unchanged)
* [ ] Reset the build number to `0` (if the version changed)
* [ ] [Re-rendered]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#rerendering-feedstocks ) with the latest `conda-smithy` (Use the phrase <code>@<space/>conda-forge-admin, please rerender</code> in a comment in this PR for automated rerendering)
* [ ] Ensured the license file is being packaged.

PR #255, opened 2023-09-29
meson on windows with clang-cl & flang; python 3.12 support

### LINKED ISSUE #164: Proposed build system changes to SciPy
Please see issue ( https://github.com/scipy/scipy/issues/13615 ) for details

### LINKED ISSUE #213: Meson integration on windows - discussion about compiler stack
Scipy has switched to meson as of scipy 1.9, and I know from conversation with @rgommers that upstream is eager to drop support for the old distutils builds.

As things stand, conda-forge is not yet able to build scipy using meson, especially on windows, and as such this would put is in a big enough pickle that we'd probably plead with upstream not to drop distutils support yet - I think it's likely that this would be heard, but at the same time, I discussed opening this issue with Ralf so that we can at least plan the necessary steps forward on conda-forge side.

The main issue being that meson relies on having a [coherent](https://github.com/conda-forge/scipy-feedstock/pull/205#issuecomment-1168450780) compiler stack for C/C++/Fortran, and the msvc+gfortran mix that was heroically hacked [together](https://github.com/conda-forge/scipy-feedstock/pull/146) by @isuruf (not having had windows-builds for scipy in conda-forge before) is not supported, and will not be.

On windows this is problematic, because we don't have a lot of fortran compiler options, or at least we _didn't_ for a long time. Possible options nowadays:
* MinGW (w/ gfortran)
  * I could imagine that this won't work for ABI reasons with the rest of the MSVC-derived conda-forge ecosystem, though notably, upstream scipy [uses](https://github.com/scipy/scipy/blob/main/.github/workflows/wheels.yml#L123-L126) a  full mingw stack.
  * Upsides: ?
  * Downsides: might be ABI-incompatible
* MSVC (or LLVM?) + flang
  * fortran support in LLVM has been a long time coming; so long, that some people have lost faith in it. However, things are progressing, and now there's experimental code generation support in LLVM 15. Indicatively, Fujitsu recently let flang loose on one of their test suites and [found](https://discourse.llvm.org/t/request-advice-on-reporting-many-errors-found-in-fortran-test-sets/65868) that 3447/8247 programs failed. There's still a lot of work to do, but it might be usable, especially because the fortran in scipy is rather old and doesn't use features from newer fortran standards. In particular, raising bugs in LLVM would probably make it possible to sort most things out by LLVM 16.
  * Upsides: LLVM is ABI-compatible with MSVC, license-compatible with redistribution & we use llvm-openmp on windows in various places already
  * Downsides: LLVM 15 (and especially flang) hasn't been built in conda-forge yet (infra changes, lack of time on my side), maturity of compiler might be unsatisfactory
* MSVC (or LLVM?) + lfortran
  * LFortran aims to be a general-purpose Fortran compiler, and is particularly putting [emphasis](https://fortran-lang.discourse.group/t/conda-toolchain-for-fortran-on-windows/3206/2) on supporting the compilation of scipy.
  * Upsides: it's in conda-forge [already](https://github.com/conda-forge/lfortran-feedstock) and actively maintained (CC @conda-forge/lfortran)
  * Downsides: support for compiling scipy is unclear, and I'm assuming meson would need to be taught to recognise LFortran (at least, it doesn't really [show up](https://cs.github.com/mesonbuild/meson?q=lfortran) in the repo; CC @eli-schwartz), ABI impact is unclear to me (but probably fine since LLVM-based), maturity of compiler might be unsatisfactory
*  MSVC + ifort
    * Ralf pointed out to me that Intel's ifort is now much more available since it moved under the oneAPI umbrella. It's [licensing](https://community.intel.com/t5/Intel-Fortran-Compiler/Intel-Fortran-2022-0-077-licensing/m-p/1352004) situation doesn't allow redistribution of the compiler though, so AFAICT we'd be in a similar situation as with nvcc - meaning we'd have to do this in the ci setup (downloading binaries). Not sure if there have been thoughts about packaging ifort for conda-forge, similar to what's happening for MKL (CC @conda-forge/intel_repack).
    * Upsides: mature compiler, interaction with MSVC/MKL works
    * Downsides: not easy to integrate with conda-forge, might need checking with Intel re:license in any case

Other issues:
* [ ] cross-compiling osx-arm64, linux-aarch64, linux-ppc64le
* [ ] [support](https://github.com/scipy/scipy/issues/17244) for alternative BLAS/LAPACK implementations in meson
* [ ] Being able to match MKL's Fortran ABI, see e.g. [here](https://github.com/conda-forge/scipy-feedstock/blob/main/recipe/bld.bat#L12-L14) and [here](https://github.com/scipy/scipy/blob/main/scipy/_build_utils/_fortran.py#L96-L102)

CC @conda-forge/core

h-vetinari:
This is a cleaned-up version of #246, taking into account learnings & patches contributed by several people (also in #252 & #253), some of which were directly upstreamed to scipy and made it into 1.11.3 already.

Requires builds from https://github.com/conda-forge/flang-feedstock/pull/28 & ~https://github.com/conda-forge/meson-feedstock/pull/89~ https://github.com/conda-forge/meson-feedstock/pull/92 which are currently only available under `llvm_rc` and `meson_dev` labels. This will hopefully improve soon.

It's likely that this will still need some work around compiler-activation, flags, etc.; but with a passing test suite in #246, we should be good for more in-depth review.

Closes #253
Closes #252 
Closes #246 
Closes #213
Closes #164

### DIFF
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -25,7 +25,7 @@ FOR /F "tokens=* USEBACKQ" %%F IN (`clang.exe -dumpversion`) DO (
 :: https://github.com/llvm/llvm-project/issues/63741
 set "FFLAGS=-D_CRT_SECURE_NO_WARNINGS -D_MT -D_DLL --target=x86_64-pc-windows-msvc -nostdlib"
 set "LDFLAGS=--target=x86_64-pc-windows-msvc -nostdlib -Xclang --dependent-lib=msvcrt -fuse-ld=lld"
-set "LDFLAGS=%LDFLAGS% -Wl,-defaultlib:%LIBRARY_PREFIX%/lib/clang/!CLANG_VER!/lib/windows/clang_rt.builtins-x86_64.lib"
+set "LDFLAGS=%LDFLAGS% -Wl,-defaultlib:%BUILD_PREFIX%/Library/lib/clang/!CLANG_VER!/lib/windows/clang_rt.builtins-x86_64.lib"
 
 :: see explanation here:
 :: https://github.com/conda-forge/scipy-feedstock/pull/253#issuecomment-1732578945
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -2,6 +2,14 @@
 
 mkdir builddir
 
+:: check if clang-cl is on path as required
+clang-cl.exe --version
+if %ERRORLEVEL% neq 0 exit 1
+
+:: set compilers to clang-cl
+set "CC=clang-cl"
+set "CXX=clang-cl"
+
 :: flang 17 still uses "temporary" name
 set "FC=flang-new"
 
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -2,6 +2,9 @@
 
 mkdir builddir
 
+:: flang 17 still uses "temporary" name
+set "FC=flang-new"
+
 :: -wnx flags mean: --wheel --no-isolation --skip-dependency-check
 %PYTHON% -m build -w -n -x ^
     -Cbuilddir=builddir ^
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -1,9 +1,0 @@
-@echo on
-setlocal enabledelayedexpansion
-
-:: copy %SRC_DIR%/base somewhere to be able to hard-reset state in build-output.bat
-mkdir .\backup
-robocopy base backup /E >nul
-:: for whatever reason, robocopy returns 1 in case of success, see
-:: https://learn.microsoft.com/en-us/troubleshoot/windows-server/backup-and-storage/return-codes-used-robocopy-utility
-if %ERRORLEVEL% neq 1 exit 1
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -10,5 +10,6 @@ set "FC=flang-new"
     -Cbuilddir=builddir ^
     -Csetup-args=-Dblas=blas ^
     -Csetup-args=-Dlapack=lapack ^
+    -Csetup-args=-Dfortran_std=none ^
     -Csetup-args=-Duse-g77-abi=true
 if %ERRORLEVEL% neq 0 (type builddir\meson-logs\meson-log.txt && exit 1)
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -1,0 +1,11 @@
+@echo on
+
+mkdir builddir
+
+:: -wnx flags mean: --wheel --no-isolation --skip-dependency-check
+%PYTHON% -m build -w -n -x ^
+    -Cbuilddir=builddir ^
+    -Csetup-args=-Dblas=blas ^
+    -Csetup-args=-Dlapack=lapack ^
+    -Csetup-args=-Duse-g77-abi=true
+if %ERRORLEVEL% neq 0 (type builddir\meson-logs\meson-log.txt && exit 1)
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -13,6 +13,13 @@ set "CXX=clang-cl"
 :: flang 17 still uses "temporary" name
 set "FC=flang-new"
 
+:: set up clang-cl correctly, see
+:: https://github.com/conda-forge/clang-win-activation-feedstock/blob/main/recipe/activate-clang_win-64.sh
+set "CPPFLAGS=-D_CRT_SECURE_NO_WARNINGS -D_MT -D_DLL --target=x86_64-pc-windows-msvc -nostdlib -Xclang --dependent-lib=msvcrt -fuse-ld=lld -fno-aligned-allocation"
+set "FFLAGS=-D_CRT_SECURE_NO_WARNINGS -D_MT -D_DLL --target=x86_64-pc-windows-msvc -nostdlib"
+set "LDFLAGS=--target=x86_64-pc-windows-msvc -nostdlib -Xclang --dependent-lib=msvcrt -fuse-ld=lld"
+set "LDFLAGS=%LDFLAGS% -Wl,-defaultlib:%LIBRARY_PREFIX%/lib/clang/17.0.0/lib/windows/clang_rt.builtins-x86_64.lib"
+
 :: -wnx flags mean: --wheel --no-isolation --skip-dependency-check
 %PYTHON% -m build -w -n -x ^
     -Cbuilddir=builddir ^
--- a/recipe/bld.bat
+++ b/recipe/bld.bat
@@ -1,4 +1,5 @@
 @echo on
+setlocal enabledelayedexpansion
 
 mkdir builddir
 
@@ -13,12 +14,18 @@ set "CXX=clang-cl"
 :: flang 17 still uses "temporary" name
 set "FC=flang-new"
 
-:: set up clang-cl correctly, see
+:: need to read clang version for path to compiler-rt
+FOR /F "tokens=* USEBACKQ" %%F IN (`clang.exe -dumpversion`) DO (
+    SET "CLANG_VER=%%F"
+)
+
+:: attempt to match flags for flang as we set them for clang-on-win, see
 :: https://github.com/conda-forge/clang-win-activation-feedstock/blob/main/recipe/activate-clang_win-64.sh
-set "CPPFLAGS=-D_CRT_SECURE_NO_WARNINGS -D_MT -D_DLL --target=x86_64-pc-windows-msvc -Xclang --dependent-lib=msvcrt -fuse-ld=lld"
+:: however, -Xflang --dependent-lib=msvcrt currently fails as an unrecognized option, see also
+:: https://github.com/llvm/llvm-project/issues/63741
 set "FFLAGS=-D_CRT_SECURE_NO_WARNINGS -D_MT -D_DLL --target=x86_64-pc-windows-msvc -nostdlib"
 set "LDFLAGS=--target=x86_64-pc-windows-msvc -nostdlib -Xclang --dependent-lib=msvcrt -fuse-ld=lld"
-set "LDFLAGS=%LDFLAGS% -Wl,-defaultlib:%LIBRARY_PREFIX%/lib/clang/17.0.0/lib/windows/clang_rt.builtins-x86_64.lib"
+set "LDFLAGS=%LDFLAGS% -Wl,-defaultlib:%LIBRARY_PREFIX%/lib/clang/!CLANG_VER!/lib/windows/clang_rt.builtins-x86_64.lib"
 
 :: see explanation here:
 :: https://github.com/conda-forge/scipy-feedstock/pull/253#issuecomment-1732578945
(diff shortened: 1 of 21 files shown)


h-vetinari:
Don't let the mood be spoiled by the pypy failure, it happens similarly on `main` and is due to https://github.com/conda-forge/pypy3.6-feedstock/issues/109. ;-)

rgommers (approved):
The recipe changes look quite good to me. I'll approve based on what I understand. The two things I can't judge are:

1. The use of labels (`conda-forge/label/llvm_rc,conda-forge/label/meson_dev,conda-forge`)
2. The Windows-specific flags in `bld.bat`

h-vetinari:
> The use of labels

It's not ideal, but both are "only" build tools, and generate no run-time dependence for scipy from a non-main channel, so I'm not really worried about that part (though of course I'm trying to get rid of these labels by finishing up [flang](https://github.com/conda-forge/flang-feedstock/pull/31) and [meson support](https://github.com/mesonbuild/meson/issues/12306)).

> The Windows-specific flags in `bld.bat`

This is one thing where I'd like to have an in-depth review. The fact that the test suite passes is already a good start, but there are many ABI-relevant pieces at play here (clang vs. clang-cl, link.exe vs. lld-link, compiler-rt vs. msvcrt vs. libcmt, flang vs. all of the above, etc.).

In particular, we might need to switch some of the flags to MSVC-syntax so that clang-cl will take them. Some of that could eventually also move to the [clang_win-64](https://github.com/conda-forge/clang-win-activation-feedstock/) activation; in any case, I'm hoping @isuruf, @xhochy et al. will be able to chime in.

h-vetinari:
No idea why meson would not pick up flang _specifically_ on python 3.12 - those things should be completely unrelated? 🤔

```
..\meson.build:82:0: ERROR: Unknown compiler(s): [['flang-new']]
The following exception(s) were encountered:
Running `flang-new --version` gave "[WinError 2] The system cannot find the file specified"
Running `flang-new -V` gave "[WinError 2] The system cannot find the file specified"
```

mattip:
@conda-forge-admin, please restart ci

mattip:
Whoops, sorry @h-vetinari. Our attempts crossed wires.

recipe/bld.bat
```diff
@@ -1,9 +1,34 @@
...
+:: set compilers to clang-cl
+set "CC=clang-cl"
+set "CXX=clang-cl"
+
+:: flang 17 still uses "temporary" name
+set "FC=flang-new"
+
+:: set up clang-cl correctly, see
+:: https://github.com/conda-forge/clang-win-activation-feedstock/blob/main/recipe/activate-clang_win-64.sh
+set "CPPFLAGS=-D_CRT_SECURE_NO_WARNINGS -D_MT -D_DLL --target=x86_64-pc-windows-msvc -Xclang --dependent-lib=msvcrt -fuse-ld=lld"
```
isuruf, line 18: This is not needed since clang-cl is being used.

recipe/bld.bat
```diff
@@ -1,9 +1,34 @@
...
+set "FFLAGS=-D_CRT_SECURE_NO_WARNINGS -D_MT -D_DLL --target=x86_64-pc-windows-msvc -nostdlib"
+set "LDFLAGS=--target=x86_64-pc-windows-msvc -nostdlib -Xclang --dependent-lib=msvcrt -fuse-ld=lld"
+set "LDFLAGS=%LDFLAGS% -Wl,-defaultlib:%LIBRARY_PREFIX%/lib/clang/17.0.0/lib/windows/clang_rt.builtins-x86_64.lib"
```
isuruf, line 21: Use `clang -dumpversion` instead of hardcoding?

recipe/bld.bat
```diff
@@ -1,9 +1,34 @@
...
+set "CC=clang-cl"
+set "CXX=clang-cl"
+
+:: flang 17 still uses "temporary" name
+set "FC=flang-new"
+
+:: set up clang-cl correctly, see
+:: https://github.com/conda-forge/clang-win-activation-feedstock/blob/main/recipe/activate-clang_win-64.sh
+set "CPPFLAGS=-D_CRT_SECURE_NO_WARNINGS -D_MT -D_DLL --target=x86_64-pc-windows-msvc -Xclang --dependent-lib=msvcrt -fuse-ld=lld"
+set "FFLAGS=-D_CRT_SECURE_NO_WARNINGS -D_MT -D_DLL --target=x86_64-pc-windows-msvc -nostdlib"
```
isuruf, line 19: You probably need `--dependent-lib=msvcrt` here as well since libraries to link to are hardcoded in compilation objects on windows.

h-vetinari: When I tried it with `-Xclang` it gets ignored, when I tried it with `-Xflang` it fails the build, see also https://github.com/llvm/llvm-project/issues/63741

recipe/bld.bat
h-vetinari: Yeah, that was a leftover from trying to follow your [request](https://github.com/conda-forge/scipy-feedstock/pull/246#issuecomment-1732692927) of matching the flags from the clang-win activation.

However, it's possible that compiler-rt is not actually necessary? I guess we'd have to try, because as a static lib it doesn't show up in the link check.

h-vetinari (commented):
Thanks for the review! Flang doesn't understand `--dependent-lib` as an option yet, so I'm not sure what we should do with that for now.

h-vetinari:
OK, so it seems we were picking up some symbols from compiler-rt after all:
```
[259/1628] Linking target scipy/special/_specfun.cp310-win_amd64.pyd
FAILED: scipy/special/_specfun.cp310-win_amd64.pyd 
"flang-new"  -Wl,/OUT:scipy/special/_specfun.cp310-win_amd64.pyd scipy/special/_specfun.cp310-win_amd64.pyd.p/meson-generated_..__specfunmodule.c.obj "-LC:/bld/scipy-split_1696193939756/_build_env/Library/lib/clang/17" "-Wl,/nologo" "-Wl,/OPT:REF" "-Wl,/DLL" "-Wl,/IMPLIB:scipy\special\_specfun.cp310-win_amd64.lib" "--target=x86_64-pc-windows-msvc" "-nostdlib" "-Xclang" "--dependent-lib=msvcrt" "-fuse-ld=lld" "-D_CRT_SECURE_NO_WARNINGS" "-D_MT" "-D_DLL" "--target=x86_64-pc-windows-msvc" "-nostdlib" "scipy\special\libspecfun.a" "scipy\lib_fortranobject.a" "%PREFIX%\libs\python310.lib" "-lFortranRuntime" "-lFortranDecimal"
lld-link: error: undefined symbol: __divdc3

>>> referenced by libspecfun.a(specfun_specfun.f.obj):(cpdsa_)
>>> referenced by libspecfun.a(specfun_specfun.f.obj):(cpdsa_)
>>> referenced by libspecfun.a(specfun_specfun.f.obj):(cfs_)
>>> referenced 318 more times
```

I couldn't find `__divdc3` for msvc anywhere (in docs online, or in a local SDK), so I think we'll have to actually stay on compiler-rt here.

recipe/bld.bat
h-vetinari: For reference, here's a [CI run](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=791496&view=logs&j=171a126d-c574-5c8c-1269-ff3b989e923d&t=623c2ba8-c4f1-582d-6c78-1c7699e2e0a2) where I had tried this in #246

h-vetinari (commented):
@isuruf, I think I've addressed all your points (short of patching upstream flang to process `--dependent-lib`...)

Do you have any remaining concerns here?

h-vetinari:
> Do you have any remaining concerns here?

Gentle ping @isuruf :)

recipe/bld.bat
```diff
@@ -1,9 +1,41 @@
...
+    SET "CLANG_VER=%%F"
+)
+
+:: attempt to match flags for flang as we set them for clang-on-win, see
+:: https://github.com/conda-forge/clang-win-activation-feedstock/blob/main/recipe/activate-clang_win-64.sh
+:: however, -Xflang --dependent-lib=msvcrt currently fails as an unrecognized option, see also
+:: https://github.com/llvm/llvm-project/issues/63741
+set "FFLAGS=-D_CRT_SECURE_NO_WARNINGS -D_MT -D_DLL --target=x86_64-pc-windows-msvc -nostdlib"
+set "LDFLAGS=--target=x86_64-pc-windows-msvc -nostdlib -Xclang --dependent-lib=msvcrt -fuse-ld=lld"
+set "LDFLAGS=%LDFLAGS% -Wl,-defaultlib:%LIBRARY_PREFIX%/lib/clang/!CLANG_VER!/lib/windows/clang_rt.builtins-x86_64.lib"
```
isuruf, line 28:
```suggestion
set "LDFLAGS=%LDFLAGS% -Wl,-defaultlib:%BUILD_PREFIX%/Library/lib/clang/!CLANG_VER!/lib/windows/clang_rt.builtins-x86_64.lib"
```

recipe/meta.yaml
```diff
@@ -199,31 +198,16 @@ outputs:
...
         - {{ pin_subpackage('scipy', exact=True) }}
-        # need to repeat host deps on windows because we have to do a full rebuild here;
-        # on unix we need a minimum as well so that installation of the wheel will work
+        # need a minimum for (re-)installation of wheel
         - python
         - pip
         - numpy
-        - libblas         # [win]
-        - libcblas        # [win]
-        - liblapack       # [win]
```
isuruf, line 221: This line is still needed.

> WARNING (scipy-tests,Lib/site-packages/scipy/integrate/_test_odeint_banded.cp311-win_amd64.pyd): Needed DSO Library/bin/liblapack.dll found in ['liblapack']
WARNING (scipy-tests,Lib/site-packages/scipy/integrate/_test_odeint_banded.cp311-win_amd64.pyd): .. but ['liblapack'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)

h-vetinari: Sure, conda cannot assume that transitive dependencies keep the ABI. Though in this case `scipy-tests` depends exactly on `scipy`, and so it's not actually a big issue. But OK.

h-vetinari:
Took the opportunity to rebase out things we don't need anymore to reduce the diff a bit:
* https://github.com/conda-forge/scipy-feedstock/pull/255/commits/05e0a5fb8c869be315bcadc0970c6ca04a6ccf99
* https://github.com/conda-forge/scipy-feedstock/pull/255/commits/1d8d247d5a284355b3d3a770a4a676895390af43 (in case support improves again)

recipe/meta.yaml
h-vetinari: Same applies to linux/osx too! 🙃

h-vetinari:
Link checks are tidy now and the windows builds are green; given that the latter was the key change here, I'm going to put this in. 

Thanks for the help everyone! What a journey it's been! 😅 🚀

traversaro:
Thanks a lot @h-vetinari for the great effort on this!

PR #262, opened 2023-12-21
SciPy 1.12.0

h-vetinari:
~Let's try to build the release candidate~

Closes #263


h-vetinari:
It looks like boost is using an unfortunate choice of template variable name on windows, which conflicts with a macro:
```
In file included from ..\scipy\_lib\boost_math\include\boost/math/policies/policy.hpp:11:
..\scipy\_lib\boost_math\include\boost/math/tools/mp.hpp(171,34): error: expected parameter declarator
  171 | template<typename L, std::size_t I>
      |                                  ^
C:\Program Files (x86)\Windows Kits\10\include\10.0.22621.0\ucrt\complex.h(63,20): note: expanded from macro 'I'
   63 | #define I          _Complex_I
      |                    ^
C:\Program Files (x86)\Windows Kits\10\include\10.0.22621.0\ucrt\complex.h(62,29): note: expanded from macro '_Complex_I'
   62 | #define _Complex_I _FCbuild(0.0F, 1.0F)
      |                             ^
```

andyfaff:
Seems that boost should probably change their macro if it shadows something further upstream?

maresb:
Only on ppc64le:

```
FAILED stats/tests/test_distributions.py::TestLaplace::test_fit_MLE_comp_optimizer[0.5-0.2] - assert (51.80214127750128 < 51.802141277501164 or False)
 +  where False = <function allclose at 0x10001f113df0>(51.80214127750128, 51.802141277501164, atol=1e-15, rtol=1e-15)
 +    where <function allclose at 0x10001f113df0> = np.allclose
```

maresb:
Looks like the RNG for Laplace is missing the seed. I might be missing something but I bet it will succeed on rerun.

https://github.com/scipy/scipy/blob/44b39e1e740795aa8cd4d69e88d6c0acb116eeac/scipy/stats/tests/test_distributions.py#L2999

maresb:
After checking the results of the build-upload job, I noticed that there was a failure in ppc64le / Python 3.12 with the same numbers as the failure above on ppc64le / Python 3.11. I wrote a note about it [here](https://github.com/scipy/scipy/pull/19937#issuecomment-1903644405). Hopefully my upstream PR resolves this for `scipy >=1.12.1`.

PR #274, opened 2024-04-20
Rebuild for numpy 2.0

h-vetinari:
This is for testing the migrator; also would need https://github.com/conda-forge/numpy-feedstock/pull/313


h-vetinari:

.azure-pipelines/azure-pipelines-linux.yml
```diff
@@ -12,10 +12,6 @@ jobs:
         CONFIG: linux_64_numpy1.22python3.10.____cpythonpython_implcpython
         UPLOAD_PACKAGES: 'True'
         DOCKER_IMAGE: quay.io/condaforge/linux-anvil-cos7-x86_64
-      linux_64_numpy1.22python3.9.____73_pypypython_implpypy:
-        CONFIG: linux_64_numpy1.22python3.9.____73_pypypython_implpypy
-        UPLOAD_PACKAGES: 'True'
-        DOCKER_IMAGE: quay.io/condaforge/linux-anvil-cos7-x86_64
```
h-vetinari, line 18: We shouldn't be dropping the pypy (or py312) builds for 1.x. AFAICT this happens because the migrator for those can only add _one_ key, not two (after the duplication here). More details in https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5790.

h-vetinari:
@conda-forge/scipy, this is looking alright for everything except PyPy, which fails in three different ways on 3 different platforms - a hang on linux, missing `__ror__` on osx (see https://github.com/scipy/scipy/issues/20472) and crashes on windows with
```
+++++++++++++++++++++++++++++++++++ Timeout +++++++++++++++++++++++++++++++++++
 WARNING: Maximum dimension of Krylov subspace exceeded prior to convergence. Try increasing KMAX.
```
Note our PyPy builds are also broken on `main`, so this is not specifically numpy 2.0 related. I'm tending towards dropping PyPy here for now. CC @mattip

mattip:
> I'm tending towards dropping PyPy here for now

Makes sense, it would be nice to have a tracking issue to point to, so if there some interest in picking up PyPy support they have something to start from.

h-vetinari:
> it would be nice to have a tracking issue to point to

Do you mean on the feedstock or in upstream scipy?

Also, I just realised that this is going to play havoc with the migration for any SciPy-dependent feedsfock. Perhaps we should keep PyPy building against 1.x for now (not just for SciPy but in the migrator)?

mattip:
> Perhaps we should keep PyPy building against 1.x

Maybe it is time to drop PyPy from future migrations. Packages that have been built should continue to work, so current users would not be suddenly cut off, and maintainers would not have to deal with the extra headache. I don't see a PyPy3.10 migration happening, and 3.9 is being dropped in the scientific python ecosystem.

h-vetinari:
> > Perhaps we should keep PyPy building against 1.x
> 
> Maybe it is time to drop PyPy from future migrations.

Those two things are equivalent in this context. Thanks for confirming.

.ci_support/win_64_numpy2.0python3.10.____cpythonpython_implpypy.yaml
```diff
@@ -15,15 +15,15 @@ libcblas:
...
+- '2.0'
 pin_run_as_build:
   python:
     min_pin: x.x
     max_pin: x.x
 python:
 - 3.10.* *_cpython
 python_impl:
-- cpython
+- pypy
```
h-vetinari, line 26: So I spent the last ~2-3 hours debugging this, made harder by the fact that the rerender changes based on random print statements or other changes I've made to my local smithy checkout (or even the `ordering:` length in the pypy migrator).

I ended up writing my findings in https://github.com/conda-forge/conda-smithy/issues/1917

h-vetinari: Now using a dev-version of smithy to work around this. If the bot does add it back, it's not a big issue as this value doesn't actually influence anything in the build.

PR #284, opened 2024-09-14
Test 1.15.0rc1

h-vetinari:
~For https://github.com/conda-forge/conda-smithy/pull/2060 & https://github.com/scipy/scipy/pull/21553~

~https://github.com/scipy/scipy/issues/21561~

Testing for 1.15.0rc1

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -271,6 +271,7 @@ outputs:
         {% set tests_to_skip = tests_to_skip + "))" %}                                                    # [aarch64]
         # on the slowest agents, these tests take more than 20min in emulation
         {% set tests_to_skip = tests_to_skip + " or test_mixed_threads_processes or test_sf_gh18506" %}   # [aarch64 or ppc64le]
+        {% set tests_to_skip = tests_to_skip + " or test_quad_vec_pool" %}                                # [aarch64 or ppc64le]
         {% set tests_to_skip = tests_to_skip + " or (test_lsq_linear and test_large_rank_deficient)" %}   # [aarch64 or ppc64le]
         {% set tests_to_skip = tests_to_skip + " or (test_iterative and test_precond_inverse)" %}         # [aarch64 or ppc64le]
         {% set tests_to_skip = tests_to_skip + " or (test_multiprocess and fft)" %}                       # [aarch64 or ppc64le]
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "1.14.1" %}
+{% set version = "1.15.0rc1" %}
 
 {% if numpy is not defined %}
 {% set numpy = "2.0.0" %}
@@ -15,20 +15,20 @@ source:
   # we can compile pythran ourselves (or not), but manually need to include
   # the submodules (not in tarball due to dear-github/dear-github#214)
   - url: https://github.com/scipy/scipy/archive/refs/tags/v{{ version }}.tar.gz
-    sha256: 233b1d5a9a7740c247e2d677e10273727abe867a415bfb4b88030b29195c495d
+    sha256: 72c3a5951d6f87f55f4210299cdcaf9af1631c8e9cb1a43ee69d6152d2439738
   # https://github.com/scipy/scipy/tree/v{{ version }}/scipy/_lib
   - folder: scipy/_lib/array_api_compat
     git_url: https://github.com/data-apis/array-api-compat.git
-    git_rev: fd22a73e19b2273035e15e8bce57f767dee8dff7
+    git_rev: 498f086568362002185b005a0a7f38ad136ca8bb
+  - folder: scipy/_lib/array_api_extra
+    git_url: https://github.com/lucascolley/array-api-extra.git
+    git_rev: 8e1c8fa42df403231d65b3690e73098cfc52674a
   - folder: scipy/_lib/boost_math
     git_url: https://github.com/boostorg/math.git
-    git_rev: a53b013c735caa98179532a32ad24d34569b9710
+    git_rev: 260f7600e95dcf87b228d46a4a37cf594df5437b
   - folder: scipy/_lib/cobyqa
     git_url: https://github.com/cobyqa/cobyqa.git
-    git_rev: 7f40b6dd54525d7aa722f9558c186d39d163af94
-  - folder: scipy/_lib/highs
-    git_url: https://github.com/scipy/highs.git
-    git_rev: 4a122958a82e67e725d08153e099efe4dad099a2
+    git_rev: 55c8e5a5d29ba5c873d87ad27bd9cffee94ae3df
   - folder: scipy/_lib/pocketfft
     git_url: https://github.com/scipy/pocketfft.git
     git_rev: 9367142748fcc9696a1c9e5a99b76ed9897c9daa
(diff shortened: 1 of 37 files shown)


h-vetinari:
Raised an issue for https://github.com/conda-forge/scipy-feedstock/commit/5eb8df4cf66361ed0e7fea8b49b87918b48e5904 / https://github.com/conda-forge/scipy-feedstock/pull/284/commits/36a062ee247d398cf80ee94a22082c38fc9f306f: https://github.com/conda-forge/conda-smithy/issues/2061

h-vetinari:
I cannot tell exactly what's going wrong here @ilayn, but note that we make it to the very last of the

<details>
<summary>import tests</summary>

https://github.com/conda-forge/scipy-feedstock/blob/07e1f9fec33af756a7f26fc69dc43529746ebaff/recipe/meta.yaml#L130-L172

</details>

In fact, we even do so twice (well, conda-build does a weird log duplication for the import tests; not 100% sure it's actually run twice). In any case, it seems like _something_ is leaving an error code (or some other dirtied resource) behind so that conda-build sees some kind of failure at the end of the import tests. It's clear as mud to me what might be the cause.

ilayn:
Whoa, that's really cryptic stuff. Regardless of the issue hats off to you for handling this chaos. I can only see one tiny sliver of hope that the failure in the azure pipelines stacktrace ends with

```
conda_build.exceptions.CondaBuildUserError: TESTS FAILED: scipy-1.14.1-py313hdf33807_0.conda
```

 https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1028042&view=logs&j=00f5923e-fdef-5026-5091-0d5a0b3d5a2c&t=f42c1d49-d161-5453-1c0e-990f827a8536&l=10989

Also the scipy version a bit above in the package list says SciPy version 1.14.1 and local. So maybe a cache needs to be invalidated? 

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1028042&view=logs&j=00f5923e-fdef-5026-5091-0d5a0b3d5a2c&t=f42c1d49-d161-5453-1c0e-990f827a8536&l=10859

h-vetinari:
Unfortunately you're fishing for red herrings here. 😅 

That's a regular message for failing tests, and the `local` is important because it means we're testing the just-built scipy (rather than an already-published one). Finally, the version is reported as old because I haven't updated the conda-side metadata in this PR.

h-vetinari:
But it is indeed unusual to fail already at the import stage without any other info. During the pytest stage we normally get better errors. I guess that's also an option here - disable the import tests and see what happens...

ilayn:
I think this is not scipy related but something else is broken in the toolchain. Looks like the .bat file is returning an error.

ilayn:
But nevermind, I already saw that it is building. I don't want to cause more work for you.

h-vetinari:
> I think this is not scipy related but something else is broken in the toolchain. Looks like the .bat file is returning an error.

Unfortunately, it's 100% guaranteed to be related to the changes that happened between 1.14 and your PR. If you want I can test SciPy main and see if it works as well, but you see a CI run further up where 1.14 passed CI today.

The fact that you see the failure in the bat script is just because that's the wrapper that executes the test suite. Unfortunately, it appears that the same kind of corruption also happens when we invoke pytest (probably during test collection, which already imports most of SciPy).

It's a frustrating kind of error to run into and to debug, but it's very real.

ilayn:
Hmm then I have no idea what it might be and more perplexing that all passing in the SciPy CI. Frustrating indeed

h-vetinari:
Good news (for your PR at least). SciPy `main` also fails in the same way...

ilayn:
I commented there but here it is less noise I think, but the MSVC job of scipy is going through but that is using ifx. Is it a Flang problem possibility ?

h-vetinari:
> Is it a Flang problem possibility ?

Always a possibility, but not my first suspicion. Could also be a VS2019 vs. 2022 issue. We'll see what the bisection yields

ilayn:
Do you mind adding the very last commit from https://github.com/scipy/scipy/pull/21553 just to test if it will fail again? Then I would know how to recreate the failing test locally.

h-vetinari:
CI retriggered, which should pick up the newest commits from that branch.

ilayn:
Thank you, fingers crossed.

ilayn:
Apologies Axel, I'm losing focus these days. I missed a few places that randomization was done in the old style. Could you retrigger this if you can please?

h-vetinari:
I mentioned this in the SciPy PR already, but just to allow you to move at your own speed: you can retrigger CI here yourself with the following magic incantation (and because the sources point to your branch, that will pick up whatever new commits are in there):

@conda-forge-admin, please restart ci

ilayn:
@conda-forge-admin, please restart ci

ilayn:
@conda-forge-admin, please restart ci

h-vetinari:
Congrats, bug squashed successfully! 😎

ilayn:
For the life of me, I don't know why this made a difference but I'll take it. Thank you for the support Axel, I hope this will make your life easier down the line a bit when it hits the road.

h-vetinari:
@conda-forge-admin, relint

h-vetinari:

h-vetinari:
4 timeouts out of 5 runs on aarch is pretty bad; seems there's a pretty severe performance drop for QEMU 9.1 on aarch; c.f. https://github.com/conda-forge/docker-images/pull/295

recipe/meta.yaml
```diff
@@ -1,34 +1,38 @@
-{% set version = "1.14.1" %}
+{% set version = "1.15.0" %}
 
+# Update the numpy value here to the most recent released version
+# when updating to a new scipy feature release
 {% if numpy is not defined %}
-{% set numpy = "2.0.0" %}
+{% set numpy = "2.2.1" %}
```
rgommers, line 6: Not sure if this mattered for previous builds or not, I suspect not since `numpy` is set centrally in conda-forge pinnings - but this change is more correct either way.

recipe/meta.yaml
```diff
@@ -1,34 +1,38 @@
...
 
+# ensure is_freethreading gets detected as a used variable
+# [is_freethreading]
+
 package:
   name: scipy-split
   version: {{ version }}
 
 source:
-  # The sdist distributed by scipy contains pre-compiled pythran sources,
```
rgommers, line 13: Note that this was not true since the switch to Meson. We should probably switch back to sdist from PyPI, so we can avoid all this manual git submodule handling?

recipe/meta.yaml
h-vetinari: It matters because we do the `<2.{N+3}` handling here, and this doesn't work with the global pinning (which now does major-only).

recipe/meta.yaml
h-vetinari: It's a bit annoying to update the hashes, but I do prefer building from source. I really don't want to pick up anything pre-cythonized or otherwise precompiled. If that's not the case anymore, I'm happy to switch back to the upstream tarballs.

h-vetinari:
Could you please cherry pick Lucas' commit from #291?

recipe/meta.yaml
rgommers: There are no pre-generated sources except for the trivial capturing of the git hash that's being built from in `version.py` (that one is necessary because it's VCS info). And that one is actually broken in the current recipe, switching to a proper sdist will fix it:
```python
>>> scipy.version.git_revision
'Unknown'
``` 

The GitHub-generated tarballs should never be used for Python packages if it can be avoided, because they're broken in multiple ways - this missing versioning info will happen for many Python packages.

recipe/meta.yaml
rgommers: The 1.14.1 package has this metadata:
```
 - numpy <2.4
 - numpy >=1.21,<3
 - numpy >=1.23.5
```

So it's already correct for previous builds (yay) - apparently `numpy` was defined in the conda-forge builds, or it would have said `<2.3`.

Anyway, this change is still helpful.

rgommers:
> Could you please cherry pick Lucas' commit from #291?

Sure, will do after letting the current run complete (it's almost done).

rgommers:
> (it's almost done).

or not, these aarch64 are crazy slow, 3h 17min and going. we may want to start skipping tests marked as slow - those are tested on native aarch64 CI machines upstream after all.

h-vetinari:
We're already only running the fast tests, but emulation is simply slow. It's not a big issue though, because we don't build so often. Just push, no need to wait for the runs to finish

PR #287, opened 2024-10-10
Rebuild with unpatched & GA build-deps

h-vetinari:
We also don't need `llvm_rc` anymore! 🥳

### DIFF
--- a/.azure-pipelines/azure-pipelines-osx.yml
+++ b/.azure-pipelines/azure-pipelines-osx.yml
@@ -5,7 +5,7 @@
 jobs:
 - job: osx
   pool:
-    vmImage: macOS-12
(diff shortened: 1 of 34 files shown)


h-vetinari:
@eli-schwartz happy to report that (after the fixes we discussed) meson 1.6.0rc1 works for scipy in conda-forge! :)

h-vetinari:
A year after #255, we're now able to build with "stock" packages. 🤩 

Things have come quite a way since then. :)

eli-schwartz:
:tada:

h-vetinari:
Yeah, flang is very noisy with unknown flags. In openblas I had to use `-w` to make it bearable. Hopefully this gets better over time.

PR #289, opened 2024-12-06
add scipy-typed package

lucascolley:
https://github.com/scipy/scipy/pull/21982#discussion_r1872409330

Checklist

### DIFF
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,4 +1,6 @@
 {% set version = "1.14.1" %}
+{% set version_parts = version.split('.') %}
+{% set next_micro_version = version_parts[0] ~ '.' ~ parts[1] ~ '.' ~ (parts[2]|int + 1) %}
 
 {% if numpy is not defined %}
 {% set numpy = "2.0.0" %}
@@ -209,7 +211,7 @@ outputs:
       run:
         - python
         - {{ pin_subpackage('scipy', exact=True) }}
-        - scipy-stubs ~={{ version }}.0
+        - scipy-stubs >={{ version }}.0,<{{ next_micro_version }}
     test:
       imports:
         - scipy
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,6 +1,6 @@
 {% set version = "1.14.1" %}
 {% set version_parts = version.split('.') %}
-{% set next_micro_version = version_parts[0] ~ '.' ~ parts[1] ~ '.' ~ (parts[2]|int + 1) %}
+{% set next_minor_version = version_parts[0] ~ '.' ~ (parts[1]|int + 1) %}
 
 {% if numpy is not defined %}
 {% set numpy = "2.0.0" %}
@@ -43,7 +43,7 @@ source:
     git_rev: a9ed9736ad52ff76ae1777922b700b13ca2bf0ae
 
 build:
-  number: 2
+  number: 3
   skip: true  # [py<310]
 
 requirements:
@@ -211,10 +211,17 @@ outputs:
       run:
         - python
         - {{ pin_subpackage('scipy', exact=True) }}
-        - scipy-stubs >={{ version }}.0,<{{ next_micro_version }}
+        - scipy-stubs >={{ version }}.0,<{{ next_minor_version }}
     test:
+      requires:
+        - basedmypy
       imports:
         - scipy
+      source_files:
+        - pyproject.toml
+        - .mypyignore
+      commands:
+        - stubtest --mypy-config-file=pyproject.toml --allowlist=.mypyignore --ignore-unused-allowlist scipy
 
   - name: scipy-tests
     script: build-output.sh   # [not win]
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -202,6 +202,18 @@ outputs:
         # ensure only one test is left (namely, our hint)
         - python -c "import sys; lines=open('testlog').readlines(); sys.exit(0 if any('======== 1 failed' in x for x in lines) else 1)"
 
+  - name: scipy-typed
+    requirements:
+      host:
+        - python
+      run:
+        - python
+        - {{ pin_subpackage('scipy', exact=True) }}
+        - scipy-stubs ~={{ version }}.0
+    test:
+      imports:
+        - scipy
+
   - name: scipy-tests
     script: build-output.sh   # [not win]
     script: build-output.bat  # [win]
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -1,6 +1,6 @@
 {% set version = "1.14.1" %}
 {% set version_parts = version.split('.') %}
-{% set next_minor_version = version_parts[0] ~ '.' ~ (parts[1]|int + 1) %}
+{% set next_minor_version = version_parts[0] ~ '.' ~ (version_parts[1]|int + 1) %}
 
 {% if numpy is not defined %}
 {% set numpy = "2.0.0" %}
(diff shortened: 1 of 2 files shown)


lucascolley:
cc @jorenham

recipe/meta.yaml
```diff
@@ -202,6 +202,18 @@ outputs:
...
         - python -c "import sys; lines=open('testlog').readlines(); sys.exit(0 if any('======== 1 failed' in x for x in lines) else 1)"
 
+  - name: scipy-typed
+    requirements:
+      host:
+        - python
+      run:
+        - python
+        - {{ pin_subpackage('scipy', exact=True) }}
+        - scipy-stubs ~={{ version }}.0
```
rgommers, line 212: Is the `~=` syntax universally supported? It's not present in the Conda or Conda-forge docs as far as I can tell, and it's always less clear than `>=x.y,<x+1` so I tend to not try using it.

lucascolley:

rgommers (commented):
Could you add the motivation to the PR description? The reason for the asymmetry of `scipy-stubs` using a run constraint of `scipy>=1.10` and this new package doing the reverse pin with exact minor version matching isn't completely obvious to me.

lucascolley:
The context is linked to in the PR description, where `pip install scipy[stubs]` was suggested.

scipy-stubs (main repo) has an optional dependency on SciPy>=1.10, but also a dependency group `scipy` which pins SciPy to the exact version. @jorenham can explain the distinction I'm sure. Perhaps a change is desired in the scipy-stubs recipe to include the exact pin?

This PR pins the exact version so that the stubs installed are compatible with the SciPy version.

lucascolley:
As for how this package is supposed to differ from `scipy-stubs` itself - `scipy-stubs` (by default) just gives you the stubs, but `scipy-typed` will give you both the stubs and `scipy`.

```
% uv add scipy-stubs
% uv pip list
Package           Version
----------------- --------
optype            0.7.3
scipy-stubs       1.14.1.5
typing-extensions 4.12.2
```

```
% pixi add scipy-stubs
% pixi ls
Package          Version    Build               Size       Kind   Source
bzip2            1.0.8      h99b78c6_7          120 KiB    conda  bzip2
ca-certificates  2024.8.30  hf0a4a13_0          154.8 KiB  conda  ca-certificates
libexpat         2.6.4      h286801f_0          63.2 KiB   conda  libexpat
libffi           3.4.2      h3422bc3_5          38.1 KiB   conda  libffi
liblzma          5.6.3      h39f12f2_1          96.8 KiB   conda  liblzma
libmpdec         4.0.0      h99b78c6_0          67.6 KiB   conda  libmpdec
libsqlite        3.47.0     hbaaea75_1          818 KiB    conda  libsqlite
libzlib          1.3.1      h8359307_2          45.3 KiB   conda  libzlib
ncurses          6.5        h7bae524_1          783.5 KiB  conda  ncurses
openssl          3.4.0      h39f12f2_0          2.8 MiB    conda  openssl
optype           0.7.3      py313h8f79df9_2     121.9 KiB  conda  optype
python           3.13.1     h4f43103_102_cp313  12.3 MiB   conda  python
python_abi       3.13       5_cp313             6.2 KiB    conda  python_abi
readline         8.2        h92ec313_1          244.5 KiB  conda  readline
scipy-stubs      1.14.1.5   pyhd8ed1ab_0        217.6 KiB  conda  scipy-stubs
tk               8.6.13     h5083fa2_1          3 MiB      conda  tk
tzdata           2024b      hc8b5060_0          119.5 KiB  conda  tzdata
```

jorenham:
> scipy-stubs (main repo) has an optional dependency on SciPy>=1.10, but also a dependency group `scipy` which pins SciPy to the exact version. @jorenham can explain the distinction I'm sure. Perhaps a change is desired in the scipy-stubs recipe to include the exact pin?

the reason I made `scipy` an optional dependency (i.e. installable as `scipy-stubs[scipy]`), is so that other libraries with *optional* `scipy` support can use the light-weight `scipy-stubs` as a required typing dependency, without having also make the `scipy` itself a required dependency.

rgommers:
Ah that optional dependency sounds like a good reason for the asymmetry. This PR is 👍🏼 then from my perspective. Unless there's conda-forge specific downsides to adding an extra output here that I don't see, I'm happy to merge it.

recipe/meta.yaml
```diff
@@ -202,6 +204,18 @@ outputs:
...
         - python -c "import sys; lines=open('testlog').readlines(); sys.exit(0 if any('======== 1 failed' in x for x in lines) else 1)"
 
+  - name: scipy-typed
+    requirements:
+      host:
+        - python
+      run:
+        - python
+        - {{ pin_subpackage('scipy', exact=True) }}
+        - scipy-stubs >={{ version }}.0,<{{ next_micro_version }}
```
h-vetinari, line 214: > This PR pins the exact version so that the stubs installed are compatible with the SciPy version.

Does `scipy-stubs` always have a matching version to scipy? In that case, we could simplify this to
```suggestion
        - scipy-stubs =={{ version }}.*
```
(and get rid of the extra jinja variables)

h-vetinari (commented):
Please rebase this on main and bump the build number

recipe/meta.yaml
```diff
@@ -202,6 +204,18 @@ outputs:
...
+    test:
+      imports:
+        - scipy
```
h-vetinari, line 217: is there something typing related we can test here? `import scipy` obviously shouldn't break, but that's also unlikely from just adding other run-dependencies.

recipe/meta.yaml
jorenham: Yea, it uses the supported `scipy` version (major.minor.patch) as a prefix. So currently it's at `1.14.1.5`, matching `scipy == 1.14.1`. The last version number (`5` in this case) is incremental and specific to `scipy-stubs`.

recipe/meta.yaml
jorenham: I suppose `stubtest` would be most appropriate in this case. That requires `basedmypy` run (`mypy` would reuslt in false positives), which is conda-forge: https://anaconda.org/conda-forge/basedmypy

recipe/meta.yaml
rgommers: Matching exactly on patch/micro version is almost certainly going to be annoying at some point (exact matches will go missing as soon as @jorenham is a bit late with a new release). It's also highly unlikely that there will be issues with patch version mismatches, since we don't add new functionality or change signatures in patch versions.

I'd go with `<next_minor_version`, so we get the latest patch version from `scipy.stubs`.

jorenham: > exact matches will go missing as soon as @jorenham is a bit late with a new release

For minor and major releases there'll be an RC period where I can even release `scipy-stubs` *before* `scipy` (as I plan to do with 1.15.0). But as for `scipy` patch releases, it'll be trickier, but not impossible. Because if I know beforehand when the scipy patch release will be published, it should be possible to release a matching `scipy-stubs` version justs before that, which should be "valid enough" if the API changes will be minimal.

rgommers: > it'll be trickier, but not impossible:

It's not 100% impossible, but it is indeed tricky - and there is no real reason to require it. I assume you do take a holiday sometimes at least, so why would we add a pin so exact that if a new `scipy-stubs` 1.xx.yy release is a day late then `scipy-types` completely fails to install?

lucascolley:
Separately, do you have any thoughts about making `pip install scipy[stubs]` (or maybe `scipy[typed]` for consistency) possible @rgommers ?

recipe/meta.yaml
jorenham: Yea you're right. And in case of patch releases `scipy-stubs` will for the most part 
already be compatible anyway.

lucascolley:

lucascolley:

lucascolley:
oops, didn't mean to close hit close!

recipe/meta.yaml
lucascolley: @h-vetinari is there a good way to include the Mypy config files from scipy-stubs so that they can be used by `stubtest` in the tests here?

h-vetinari: In principle, `scipy-stubs` could ship the mypy files, but I have to wonder why we're adding this output to scipy at all. It seems like it would fit much better as an additional output to `scipy-stubs` (not just due to the version relationship, but also because then it could easily make use of the mypy config files).

lucascolley: I think I agree actually, let me make a PR over there.

rgommers:
> Separately, do you have any thoughts about making `pip install scipy[stubs]` (or maybe `scipy[typed]` for consistency) possible @rgommers ?

I don't mind it, since it makes `scipy-stubs` more discoverable. I don't really like to recommend extra's for users though (it's more a developer-facing feature that keeps track of optional/dev dependencies for us), since the UX is kinda bad and if users know they want the type stubs then installing `scipy scipy-stubs` is just as easy as `scipy[stubs]`.

Implementing the same version pinning strategy is a bit more annoying in `pyproject.toml`, since you cannot template but must hardcode the version numbers. Not a real problem, but then the version number at the top needs a comment like this:
```toml
# when updating, also update the `stubs` version constraint further down.
```

lucascolley:
continued in https://github.com/conda-forge/scipy-stubs-feedstock/pull/5

PR #300, opened 2025-05-22
scipy v1.16.0

h-vetinari:
Testing the release [candidate](https://github.com/scipy/scipy/releases/tag/v1.16.0rc1), which has a bunch of bigger-than-usual changes.


rgommers:
Note that using `-Duse-system-libraries` is fine as a test exercise, but we should not use it in this feedstock. It increases the number of runtime dependencies and potential failure modes at build and runtime, for no upside. The build option is primarily useful for distros/companies that must unvendor because policies.

h-vetinari:
> The build option is primarily useful for distros/companies that must unvendor because policies.

Conda-forge unvendors because of policy too... It still happens because sometimes it's too much effort to manually patch, but ignoring an explicit unvendoring option is pretty much unheard of.

> It increases the number of runtime dependencies and potential failure modes at build and runtime, for no upside.

I don't agree here; if scipy and qhull are both used in the same environment, we avoid duplicate copies of the same thing (right now it's just qhull, but the point holds for all libraries that can eventually be unvendored).

And we run the full test suite, so any thing that's passing here is as functional as something that passes upstream CI (and now that scipy provides this option, I'd argue that it's in the interest of scipy to ensure that it works, which we would be ensuring here through testing and use).

h-vetinari:
> For example, a new version of a dependency doesn't re-run the test suite of SciPy, so can easily break already-released SciPy versions.

That's handled by appropriate pinning and run-exports. qhull currently run-exports `qhull >=2020.2,<2020.3.0a0`, so no new release would affect scipy. boost is used as header-only, so the problem doesn't even arise.

> Other example: at build time SciPy looks for specific versions (currently boost.math 1.88 and qhull 2020.2/8.0.2). So new releases of those packages may break this feedstock.

SciPy can of course enforce expectations on the third-party dependencies at build time. That's like any other dependency constraint (morally speaking like those in `pyproject.toml`) that we need to match, and we can easily do so.

> I'm 100% certain that using this option will make the `scipy` package more fragile in conda-forge. There is no problem today with either `boost.math` or `qhull`, so it's much better not to change anything here.

I disagree.

rgommers:
> That's handled by appropriate pinning and run-exports. qhull currently run-exports `qhull >=2020.2,<2020.3.0a0`, so no new release would affect already-published scipy builds.

Well, then you have a strict pin in the dependencies, that's an equally large problem. Build a new version of something else that depends on `qhull` and you now have created an install-time conflict between `scipy` and that other package.

rgommers:
That's still going to break existing environments during the migration, as well as cause more work (every PR is work) and churn for users (more artifacts to download and keep in cache). Migrators are helpful to _mitigate_ this problem, but they don't avoid it.

You have to actually have a compelling _benefit_ to want to use this new build option. Right now you have not identified one. It's just more complexity in the recipe, more rebuilds, and more downsides for users.

Such benefits could be:
- Makes maintenance easier. E.g., can drop patching because you already were unvendoring for some reason - this does occur sometimes, e.g. with HiGHS in a monorepo setup we've had a user (Google) with a real-world issue.
- Makes builds substantially faster
- Makes binary size substantially smaller
- It's a security-sensitive component that needs regular patching, and you want to avoid having to think about that in the scipy feedstock

None of these apply to `qhull` or `boost.math`. Once we add HiGHS as a supported option, it may apply there (it's a quite large C++ library), so then it may make sense to try. It's always a tradeoff, and you have to pick the component based on that tradeoff. Being dogmatic about it and unvendoring "just because" is a bad idea.

h-vetinari:
> That's still going to break existing environments during the migration

No, that's the whole point about the migration infrastructure; things don't break either the before (if not all packages in the environment have been recompiled yet; the old version gets pulled in) or after (when everything has become available; now the new version gets pulled in).

> You have to actually have a compelling _benefit_ to want to use this new build option. Right now you have not identified one.

Deduplicating qhull is a benefit. Validating that system library integration works (and keeps working going forward) is a benefit.

> It's always a tradeoff, and you have to pick the component based on that tradeoff. Being dogmatic about it and unvendoring "just because" is a bad idea.

I'm happy to make pragmatic choices, but none of the things you point out are convincing as obstacles to trying out whether the unvendoring works at scale. We deal with those exact same patterns a hundred times over in c-f all the time.

h-vetinari:
> a new version of a dependency [...] can easily break already-released SciPy versions.

> then you have a strict pin in the dependencies, that's an equally large problem.

> That's still going to break existing environments during the migration

> This is really bad actually

You're not doing your argument any favours by drastically overstating the supposed problems and/or not being familiar with how these things work in conda-forge (e.g. that migrations supposedly break something). It is in conda-forge's DNA to prefer rebuilds over vendoring things, and we know how to deal with that, because we do it literally every day.

In this case it's also an almost entirely hypothetical concern, because qhull has had three releases in the last 10 years, and even with migrating every time it'd be a completely negligible increase in overall builds & downloads.

> I've given you not one but multiple good reasons

Perhaps they sound convincing to you, but all I see so far is unfounded FUD. Perhaps you can restate the problems you expect in a more neutral way, then we can find out together if they apply, and to what degree (and whether that's worth vendoring). To respond to your last point: that matplotlib and scipy would have to be built against the same qhull is really not a constraint that matters in practice (and again, we could rebuild missing combinations easily, were that ever to be necessary).

rgommers:
> Perhaps you can restate the problems you expect in a more neutral way, then we can find out together if they apply

I've been doing all the concrete example-giving and investigation in binary size here, and coming up with concrete scenarios. I also designed the build option you're trying to use here, so I know what I want from it and what it is and isn't for.  Can we please turn this around? I would like you to give compelling reasons to go _against_ the upstream advice on usage. 

Also, can you please answer the following question. SciPy has a lot more vendored dependencies (xref https://github.com/scipy/scipy/issues/21232), so the number of libraries behind this build option may grow to O(10). For any of them, is there any reason you would not want to unvendor them? Or is your intention to just always unvendor everything no matter what, independent of maintenance status of the library in question, its binary size, whether it's a strict pin or not, etc.? 


> To respond to your last point: that matplotlib and scipy would have to be built against the same qhull is really not a constraint that matters in practice

Only if qhull never does another release. If it does, it _will_ break combinations of versions that work today. `qhull` is indeed developed very slowly, and is therefore lower-stakes, but you'll be making the exact same argument for a library that has a release every year or twice a year I suspect.

I honestly don't know why you would say that this doesn't matter in practice. It happens regularly that people pin either library due to incompatibilities (e.g., new `matplotlib` release broke doc build, introduce a `<3.new` pin to unbreak CI, investigate in the future). It's all resolvable of course, but it's also all extra work and constraints. And I also then expect questions like "hey this works when installing from PyPI, but not from conda-forge". 

> and again, we could rebuild missing combinations easily, were that ever to be necessary).

That's not what migrators do by default, so it's a bunch of extra work that no one is going to do proactively. Going back several releases may also run into new build issues, that won't automatically work (e.g., new BLAS version pulled in breaks some tests). So in practice, I'd say that the versions in question will remain incompatible.

rgommers:
> Please think about it as more concrete statements like "`environment.yml` files that used to resolve may no longer do so".

Let me retract this bit. @h-vetinari and I had a call and talked through this. The resolver should in principle always be able to resolve this, and while it could fail in corner cases (e.g., due to the problem being resolved becoming harder or a resolver bug) this is quite rare and less common today than it was in the past. Apologies for the noise on this one.

rgommers:
The rest of the things I brought up are still valid I think, for the SciPy case and the libraries we're talking about. The tl;dr is:

- There is no significant concrete benefit to this,
- The "range narrowing" (e.g. new SciPy versions and old Matplotlib versions becoming incompatible) is undesirable,
- The extra maintenance effort and rebuilds are undesirable,
- It's not possible to run a migrator without either applying or removing a patch, since SciPy checks for exact versions,

So my preferred solution is simple: do nothing, and treat 1.16.0 just like 1.15.0. 

This particular topic can always be revisit in the future, if and when this equation changes (e.g., SciPy starts supporting version ranges, or adds a new library to `-Duse-system-libraries` that will have a larger benefit to unvendor).

jorenham:
friendly bump

lucascolley:
is there value in getting https://github.com/conda-forge/scipy-feedstock/pull/301 merged first?

h-vetinari:
> is there value in getting #301 merged first?

Feel free to open a PR based on that. If you can get it to green, we can merge (timing with this PR is secondary, I can create an 1.15 branch otherwise).

lucascolley:
okay good to know that they can be merged in any order 👍

I'd be happy to try to get gh-301 if you have some pointers for what to look at

h-vetinari:
Your guess is as good as mine for now. :)

Just digging into the (build) failures and trying to figure out what's going on. Given that 1.16 was on the horizon, 1.15.3 was not a very high priority for me (given the unclear effort necessary to fix).

lucascolley:
one would hope that some of the fixes here might also be fixes for over there, so maybe I'll give it a go once you've figured this one out :) let me know if I can help, though

lucascolley:
nice stuff @h-vetinari !

